PR #4, opened 2017-08-31
Fixes

Niverton:
Hey, I really like the simplicity of this DM, but I found a few bugs when using it. Decided to see if I could fix them.
The first one is a quick one, due to a typo the username wasn't getting properly copied from the save file. Fixed it and switched to strcpy for lisibility.
The second one is initialising the user's groups when logging, without that the user session is created with only the default group.

### DIFF
--- a/src/login.c
+++ b/src/login.c
@@ -206,26 +206,25 @@ const char* de_command, enum deserv_t display_server)
 		return 1;
 	}
 
-    /* Initialise user groups */
-        /* Get pwd structure for the user to get his group id */
-    struct passwd* pw = getpwnam(username);
-    if (!pw) 
-    {
-        error_print(strerror(errno));
+	/* Initialise user groups */
+	/* Get pwd structure for the user to get his group id */
+	struct passwd* pw = getpwnam(username);
 
+	if(!pw)
+	{
+		error_print(strerror(errno));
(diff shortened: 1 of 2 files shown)


cylgom:
Thank you for your contribution! This save file bug is a shame really, I'm sorry it caused you trouble.
For the user groups it explains a lot of things... I'm glad you figured that out ^^. Your PR looks good to me, could you run astyle with [this config](https://pastebin.com/50yhMFnb) on your modifications before I merge? Thanks again!

cylgom:
I just noticed you created a new "passwd" pointer named "pw" ; I already use one called "pwd". Could you fix that ? (or maybe there is something I don't understand ?)

Niverton:
You're welcome!
I fixed the styling. The `pw` pointer is a pointer to the structure returned by `getpwnam`, needed to get the group id of the user's main group. It's necessary because for some reason you're forced to provide a group id to `initgroups`, the function that reads groups from /etc/groups and assign the user's to the session we're creating. You don't want to assign a random group to the user (during my testing I tried `initgroups(username, 0)` and ended up with my user in the `root` group), and the easiest way I found to get a group the user is in is using this password structure.
There might be better ways to do it, but I'm a noob regarding the authentication API :)

PR #19, opened 2018-02-01
added cc and make as dependencies in the README

dglanzman:
Base set of packages for common Linux distros (such as Ubuntu) don't come with cc or make by default. Maybe they should be added as dependencies in the readme.


cylgom:
This LGTM, thank you :)

PR #23, opened 2018-04-02
don't assume home is in /home/$USER

djpohly:
Adjustment of original fix by @abdullahibnnadjo


cylgom:
Thank you very much  :)

PR #26, opened 2018-04-02
navigate fields with tab/backtab

### LINKED ISSUE #15: Use Tab to change textfields
Could you add the possibility to change the textfields by using the tab key?

djpohly:
Fixes #15.


cylgom:
I am somewhat divided over this one. But I already assumed arrow keys had nothing to do in passwords so...

PR #28, opened 2018-04-05
assert LY_CFG_SAVE

arwn:
Throw an error instead of segfault when /etc/ly doesn't exist.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -1,6 +1,7 @@
 #define _XOPEN_SOURCE
 
 /* stdlib */
+#include <assert.h>
 #include <string.h>
 #include <stdlib.h>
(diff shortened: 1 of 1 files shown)


djpohly:
Does this actually work as written?  It looks like the patch asserts that the filename string `LY_CFG_SAVE` is true (i.e. not NULL) rather than the opened file handle.

It might be better just to skip saving if the file is not found or cannot be opened rather than exiting the program (whether by segfault or by abort), so that the user still gets to log in.

cylgom:
As djpohly said : this won't work, you have to check the file handle, but thank you for taking the time to work on this issue :)

Issue #14, opened 2017-12-26
Backspace changing textfields


L0ric0:
When one deletes the hole password field, the courser jumps to the username field.

oxodao:
That's intended, that's how ncurses works

cylgom:
I'll take a look, but oxodao is right : this is the default behaviour of the library I use.

AlexKornitzer:
@cylgom @oxodao, any chance that you can merge @djpohly's changes or are there reasons you have not?

oxodao:
I do not have any access to the repo so I won't merge it, and cylgom is quite busy. I will try to contact him today.

AlexKornitzer:
Ah awesome, many thanks, as they appear to fix a few issues I have with what is otherwise an awesome bit of code :)

PR #29, opened 2018-04-17
add bash/Emacs-like ctrl-key navigation

djpohly:
Also fixes backspace going back to EOL.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -22,8 +22,7 @@
 #include "desktop.h"
 #include "ncui.h"
 
-#define KEY_ENTER_ASCII 10
-#define KEY_BACKSPACE_ASCII 127
+#define KEY_CTRL(c) ((c) & 0x1f)
 
 int main(void)
 {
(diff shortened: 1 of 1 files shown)


oxodao:
AFAIK backspace going back to EOL is intended..

djpohly:
It seemed intended but misguided... I can't think of any other UI where pressing backspace in the middle of a field deletes one character and then jumps to the end of that same field.  I suspect this was a hack to mitigate issue #14 , and it was just never tested with the cursor in the middle of a field.

Combine with #26 and #27 for some very normal textfield behavior.

oxodao:
Oh, I thought you were talking about the fact that when you press backspace on the last character of a field it goes into the previous one, because that's how ncurses is.
What you are saying is weird in fact we (Cylgom and I) never encountered it as far as I know, so nice that you fixed it :)

djpohly:
I'd only suggest this behavior since it matches that found in shells, which presumably is something that ly's target audience is accustomed to. :)  (I'm not an emacs person myself.)

At least take ^U to erase the current field, as that's part of the default TTY line discipline [on a number of systems](https://www.in-ulm.de/~mascheck/various/stty/).

oxodao:
`At least take ^U to erase the current field`
Learning things everyday

Issue #38, opened 2018-07-05
Service Message Popups


L0ric0:
If tlp or plymouth are not installed Service messages will still popup over ly.

I will look for a solution myself but as I don't have any experience in c it could take time.

cylgom:
Actually this can get really tricky, we had a hard time with Oxodao trying to fix this, and the only solution we could find is described in the readme. Another working option could be to edit you kernel parameters.
Can you tell me if the first method works for you?

L0ric0:
The method described in the readme doesn't work for me or I would have to specify each of these services. 

As for the kernel parameters, I can suppress them withe the quiet option, but this also suppresses errors and I haven't found one that lets me redirect messages to a specific tty or file.

Is it possible to catch all input to the tty and redirect it to another?
I will see if I can find stuff in the agetty source as it and other programs like it have to have methods to only let errors show errors.

cylgom:
@RobertZenz this easy to do, but not quite a clean approach.
@L0ric0 you have to specify the last service launching on your computer that gives error messages.

RobertZenz:
No, it isn't, but it would be a quick and easy workaround. Especially when made optional.

L0ric0:
@cylgom Yes I know but systemd works asynchronous where it can and so I have multiple services that can be the last to launch. But I have found that you can define a target in the After section of a service. So adding:

After=multi-user.target 

to the service file will launch ly only when all services of multi-user.target are started circumventing the issue of knowing the last service to start.

PR #39, opened 2018-07-21
Disable service popup messages

L0ric0:
This change to ly.service resolves the problem with service messages
poping up over ly. #38 

It should work with all systemd default configurations but was only tested on Archlinux.

### DIFF
--- a/ly.service
+++ b/ly.service
@@ -1,6 +1,6 @@
 [Unit]
 Description=ly ncurses display manager
-After=systemd-user-sessions.service
+After=systemd-user-sessions.service plymouth-quit-wait.service
 After=getty@tty2.service
 After=multi-user.target
 
--- a/ly.service
+++ b/ly.service
@@ -1,7 +1,8 @@
 [Unit]
 Description=ly ncurses display manager
-After=systemd-user-sessions.service plymouth-quit-wait.service
+After=systemd-user-sessions.service
 After=getty@tty2.service
+After=multi-user.target
 
 [Service]
 Type=simple


oxodao:
If it really works, you're my new god.
I've been searching this for ages (Even before it was released, with the first builds)

L0ric0:
@oxodao can you try starting ly with getty? For that you have to enable the getty on tty2 and use a dropin file like this:
``[Service]
ExecStart=
ExecStart=-/sbin/agetty -n --delay 1 -l /usr/bin/ly %I $TERM
``

oxodao:
I had a problem with this method on my laptop where it hanged for a while on a black screen before actually letting ly pop up. I'll try this asap

L0ric0:
you can try removing the --delay 1 (this lets agetty wait a second before starting ly) but that led to a blank screen an the f1 an f2 messages on the top.

cylgom:
This will be addressed the dirty way in next version, with a configurable auto-refresh (which was implemented for another purpose, but you can very well use it *only* to clear the screen)

s-ayush:
So, I tried the 'dirty way' with `force-update` and `blank-box` and `animate` (as recommended by @cylgom in #79), and it did not work. I also tried adding `After=multi-user.target` (by @L0ric0 in #38)  which did not work either, I finally set `Type=idle` instead of `Type=simple` (@bl33pbl0p here) which did work!

I don't know _why_ it worked, I just thought I would drop this here nonetheless.

bl33pbl0p:
@s-ayush because Type=idle tells systemd to wait until all jobs that active are finished before starting your unit (at which point hopefully nothing writes to the console and output is not interleaved).

PR #45, opened 2018-08-10
Ignore keystrokes that are not associated with a keysym

### LINKED ISSUE #21: stucks when power plug insertion ACPI event happens
The following string is printed on the screen and then ly freezes.
[ 9910.556531 ] acpi INT3400:00: Unsupported event [ 0x86 ]

Then the only way to make ly work again is to restart the relative unit via another tty.
Other events also write on the screen but doesn't affect ly.

qub1750ul:
The system maps these to a keysym with value 0, causing ly to terminate.
With this fix it shouldn't happen anymore

Fixes #21

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -91,7 +91,7 @@ int main(void)
 	set_current_field(form.form, form.fields[6]);
 	form_driver(form.form, REQ_END_LINE);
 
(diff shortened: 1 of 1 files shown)


cylgom:
Thank you :)

PR #55, opened 2018-09-13
Configurable tty number

d5244af6a0:
The proposed changes allow to set the tty number through the environment variable LY_CONSOLE_TTY. Constant definition in the code still remain workable.


cylgom:
Given the text-file configuration system available in the upcoming version, this is no longer needed, but thank you for taking the time to do this.

PR #57, opened 2018-09-22
Redirect X session messages to log.

zackarymoore:


cylgom:
Added in upcoming version, thank you for this!

PR #67, opened 2018-10-05
Define feature macros only on Linux

liweitianux:
In preparation for the BSD support.

See also my relevant pull request to the termbox-next at:
https://github.com/cylgom/termbox-next/pull/1


cylgom:
Thank you for all your contributions and explanations :)

PR #70, opened 2018-10-05
Use sysconf(_SC_HOST_NAME_MAX) to determine maximum hostname length

liweitianux:
The `HOST_NAME_MAX` macro is not available on BSD, and the standardized
way is to call `sysconf(_SC_HOST_NAME_MAX)` to determine it, which is
the actual value that the OS is supported.

### DIFF
--- a/src/util.c
+++ b/src/util.c
@@ -24,12 +24,24 @@ void hostname(char** out)
 {
 	struct addrinfo hints;
 	struct addrinfo* info;
-	char hostname[HOST_NAME_MAX + 1];
+	char* hostname;
 	char* dot;
+	int host_name_max;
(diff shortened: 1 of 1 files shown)


cylgom:
Oh well I didn't know about this :O The only thing that makes me doubt here is the dynamic memory allocation, but for the sake of simplicity I think it is perfectly acceptable.

liweitianux:
@cylgom Thanks for reviewing and merging this.

Actually, I'm running into a problem on DragonFly BSD with `getaddrinfo()` here, which seems to fail and then cause `freeaddrinfo()` panic the program.  I have to look into further and propose a patch later.

Cheers

PR #73, opened 2018-10-05
util: Fix gethostname() and freeaddrinfo() calls

liweitianux:
I was passing the wrong size to the `gethostname()` call because the `hostname` is now dynamically allocated and cannot use `sizeof` to get its size.  This mistake causes that the obtained hostname is empty.

The `freeaddrinfo()` was placed in the wrong place.  Fix it.

Fallback to `_POSIX_HOST_NAME_MAX` if `sysconf()` fails.

Always null-terminate the `hostname`.

(Edited to match my updated patch.)

### DIFF
--- a/src/util.c
+++ b/src/util.c
@@ -16,14 +16,15 @@
 #include <sys/ioctl.h>
 #include <linux/vt.h>
 
-char* hostname_backup;
+static char* hostname_backup = NULL;
 
 void hostname(char** out)
 {
-	struct addrinfo hints;
-	struct addrinfo* info;
-	char* hostname;
-	int result;
+	if (hostname_backup != NULL)
+	{
+		*out = hostname_backup;
+		return;
+	}
 
 	int maxlen = sysconf(_SC_HOST_NAME_MAX);
 	if (maxlen < 0)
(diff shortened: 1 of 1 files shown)


src/util.c
```diff
@@ -49,16 +53,16 @@ void hostname(char** out)
 
 	if ((result == 0) && (info != NULL))
 	{
-		dot = strchr(info->ai_canonname, '.');
+		char* dot = strchr(info->ai_canonname, '.');
```
liweitianux: Do we need to check the return value of `dot`?

liweitianux:
@cylgom Thanks for reviewing.  I'll fix it later (maybe tomorrow).

One question: is there a specific reason that you prefer `type* v` instead of the more common style `type *v`? The latter should be better when you have to write something like `type *v1, *v2`.  Thanks.

Cheers,

cylgom:
That's the one eternal schism of C programmers :) There are arguments on  each side though:

`type *v`
- makes `type *v1, *v2;` more readable and appear logical
- makes function pointers look good but less readable `void (*fun_ptr)(int) = &fun;`
- makes pointer-returning function pointers less readable `int *(*fun_ptr)(int) = &fun;`
- makes it look like a pointer is not a type and thus does not reflect the way compilers work

`type* v`
- makes `type* v1, *v2;` inconsistent, obscure and completely unreadable.
- makes function pointers look bad but more readable `void (* fun_ptr)(int) = &fun;`
- makes pointer-returning function pointers more readable `int* (* fun_ptr)(int) = &fun;`
- makes it look like a pointer is a type and thus reflects the way compilers work

Actually I used to be like you and use the first spacing convention. But one day I realized `type *v1, *v2;` is very little used in practice, precisely because it's counterintuitive to most people. Carefully looking at the arguments for and against each camp, we can clearly see this is the most important one. So `type *v1, *v2;` looks a lot better, granted, but I never write such declarations, so I changed my mind and started using `type* v;`, because all the other minor arguments are in favor of this spelling.

But to me this debate confirms we need a better-designed replacement for the C language.

liweitianux:
Thanks for the detailed explanation. Since you adopt the style, then I'll follow yours.

On the other hand, I think the `hostname` obtained via `gethostname()` is sufficient, so there is no need to bother with the `getaddrinfo()` and simplify this code a lot.  I had a look at LightDM, and it only uses `gethostname()`.  What do you think?

Cheers.

cylgom:
I just tested a few things and it looks like you're right. Let's remove getaddrinfo().

liweitianux:
Hi @cylgom , I've pushed a new commit here to simplify the `hostname()` function.  Please see the commit message for some details.

Cheers.

cylgom:
Thanks a lot!

PR #77, opened 2018-10-07
login: No need to set MAIL

liweitianux:
Propose to fix https://github.com/cylgom/ly/issues/75.

See https://github.com/cylgom/ly/issues/75#issuecomment-427627825 for more details.


cylgom:
I think this is set by login from linux-utils. But I prefer Ly to compile with musl than to support something other display managers don't

PR #78, opened 2018-10-07
Fix ioctl issues on DragonFly BSD and FreeBSD

liweitianux:
* Include the correct headers on DragonFly BSD and FreeBSD.

* Use UTF-8 characters to draw the border on Linux.

* Use ioctl `KDGETLED` to get the keyboard NumLock/CapsLock states on BSD, while keep using `KDGKBLED` ioctl on Linux (see the `ioctl_console(2)` man page for more details).

* Use system macros instead of hard-coded magic numbers to extract the NumLock/CapsLock states, i.e., `LED_NUM` and `LED_CAP` on BSD, `K_NUMLOCK` and `K_CAPSLOCK` on Linux.

NOTE:
This patch only allows 'ly' to build fine on DragonFly BSD and FreeBSD (other *BSD not tested), and more works are needed to make 'ly' working fine on BSD.

### DIFF
--- a/src/draw.c
+++ b/src/draw.c
@@ -10,15 +10,20 @@
 #include <stdlib.h>
 #include <sys/time.h>
 #include <sys/ioctl.h>
-#include <linux/kd.h> 
-#include <linux/vt.h>
 #include <stdio.h>
 #include <unistd.h>
+#if defined(__DragonFly__) || defined(__FreeBSD__)
+#  include <sys/kbio.h>
+#else  /* assume Linux */
+#  include <linux/kd.h>
+#endif
 
+#if defined(__linux__)
 // border chars: ┌ └ ┐ ┘ ─ ─ │ │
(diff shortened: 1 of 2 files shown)


src/draw.c
```diff
@@ -216,14 +221,13 @@ void draw_lock_state()
 	}
 
 	int fd = fileno(console);
-	char ret;
-
-	ioctl(fd, KDGKBLED, &ret);
+	char led;
+	ioctl(fd, KDGETLED, &led);
```
cylgom: If I'm not mistaken this only gets the state of the LEDs, which can be wrong depending on reboots, keyboards etc. We need to use KDGKBLED which really represents the state of the key in the system. We can keep LED_NUM and LED_CAP though.

liweitianux: You're right that we should use the `KDGKBLED` ioctl on Linux.  See my new commit to this PR and the message there.

PR #85, opened 2018-11-24
italian language

termgod:
added new it.ini file in the lang directory


RobertZenz:
You should rename the pull request, description and the commit to only reflect the changes which you've done and remove the superfluous (uninteresting) parts.

Or to put it a little bit different, in 99.9% of the projects nobody cares whether you're new or not. They only care about your contributions, and these should be as tidy and streamlined as possible.

drozdowsky:
@termgod I have used your patch in this fork: https://github.com/drozdowsky/ly-void
You are listed as author, if you have anything against it, pm me.
Unfortunately I could not (I don't know how to properly) cherry-pick your commit.

termgod:
no problem! I'm also a Void user, so I'm glad it was useful!

‐‐‐‐‐‐‐ Original Message ‐‐‐‐‐‐‐

PR #96, opened 2019-03-16
Added Tab Cycle

cortiz:
Adds Tab Cycle to the inputs.


oxodao:
LGTM
Perhaps @cylgom you should add a config for the o/* as password mask. I'd rather have the later

PR #97, opened 2019-03-27
Fixed sourcing of .xsession

16wardcj:
Hi,
I found a trivial bug in xsetup.sh where ~/.xsession wasn't being sourced.

PS. This is my first pull request so if I messed up somehow please let me know :)


16wardcj:
Realised that ~/.xsession is supposed to be executable

PR #108, opened 2019-06-27
Customizations

HavenMKing:
Added support for 256 colors (before realizing that tty only supports the 8). Looks really pretty when run under an X server that can support it, so I was reluctant to remove it.

Also added the ability to reposition the sign-in box, by percentage of the screen.

Next steps would be adding the ability for users to specify the box_chars, including other Unicode symbols if possible.

### DIFF
--- a/src/config.c
+++ b/src/config.c
@@ -38,6 +38,18 @@ static void config_handle_u8(void* data, char** pars, const int pars_count)
 	}
 }
 
+static void config_handle_hex(void* data, char** pars, const int pars_count)
+{
+	if (strcmp(*pars, "") == 0)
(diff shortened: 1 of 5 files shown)


cylgom:
Very nice, but it looks like some variables renaming and formatting changes happened :) Please make sure to keep the original variables names and coding style.

I don't understand the point of having 256 colors though. Ly can't be used from a desktop terminal emulator anyway as it doesn't support xephyr, so we should probably keep the color mode to 8.

I you can fix these 2 things I'll review and merge.

ArchKubi:
256 colors how we use this

PR #115, opened 2019-08-13
Doom fire bug note

CodeOhms:
It would be handy for other users to know this, instead of having to search through the issues list for a solution.


cylgom:
Bug fixed in 4ba0725, thank you for urging me to address it

PR #123, opened 2019-09-19
i3wm - hide xorg start messages

raphaunix:
Hi!
I use [i3wm](https://i3wm.org/) and I really like **[ly DM](https://github.com/cylgom/ly)**, thanks for sharing.
However, xorg's start output gives the impression that something is wrong even though it is not. 
My proposal here is just to suppress it:

From:
```c
	config.x_cmd = strdup("/usr/bin/X");
```
To:
```c
	config.x_cmd = strdup("/usr/bin/X -keeptty > ~/.xorg.log 2>&1");
```

Thank you! :)


cylgom:
Hi, this is already configurable in /etc/ly/config.ini, but thank you for your interest :)

PR #126, opened 2019-10-04
cosmetic changes + enter behavior

stlaz:
Hello,

I enjoy using ly as my DM so I thought I'll see if there's something I could do to contribute.

With this patch I moved processing the key events in the main loop from if-elses to switch, which feels more natural and is perhaps better readable. That changed the behavior of "Enter" a bit from "move one line lower" to "submit password", but I guess that's not terrible (I now move the cursor to the password field on auth fail)

I also wrapped the pam actions into pam_do() function so that the clean up happens in one spot.

See if you like any of this, I haven't coded in C for quite some time, but hopefully it's not all terrible :)

### DIFF
--- a/src/config.c
+++ b/src/config.c
@@ -263,7 +263,7 @@ void config_defaults()
 	config.blank_box = true;
 	config.blank_password = false;
 	config.console_dev = strdup("/dev/console");
-	config.default_input = 2;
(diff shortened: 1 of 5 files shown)


cylgom:
I really like the pam wrapper and the changes are very good overall IMHO. I don't see any reason not to merge this ; thanks again :)

PR #127, opened 2019-10-04
print hostname in info lane on start

stlaz:
hostname was set in the info_line and unset right after that

----

partially fixes https://github.com/cylgom/ly/issues/114


cylgom:
Woops, thanks for the fix :)

Issue #121, opened 2019-09-02
How to handle login session correctly?


jkrcma:
When I use `ly` to log in and start X for me, I can't see my session in `w` or `who` command outputs. It brings problems together with `policykit` where all the default rules concerning suspend or hibernate are bound to local sessions only.

I wasn't figure out whether I am able to force local login session through `xsetup.sh` or PAM. Can you help?

inaugur:
Same issue. Login with sddm and ly makes different situations. Here my outputs (archlinux):

`sddm; w`
23:02:41 up 37 min,  1 user,  load average: 0,96, 1,04, 0,94
USER     TTY        LOGIN@   IDLE   JCPU   PCPU WHAT
myusername   tty1      22:25   37:44   1:10  34.90s awesome

`sddm; who`
myusername   tty1         2019-09-18 22:25 (:0)

`ly; w`
23:04:46 up 0 min,  0 users,  load average: 0,37, 0,09, 0,03
USER     TTY        LOGIN@   IDLE   JCPU   PCPU WHAT

`ly; who`
it prints nothing

stlaz:
This is basically a feature request on adding utmp handling for login accounting, see `man 3 getutxent`

evgenux:
In 0.4.0 issue still exist. Command `w` returns nothing. Due to the problem range of the other programs doesn't work correctly.

stlaz:
@jkrcma @evgenux I opened a PR which should fix this for you, please see whether it does - https://github.com/cylgom/ly/pull/133

evgenux:
Thank you a lot, @stlaz! With #133 login session starts well on my pc. Even my clipmanager, which depends somehow on it, functionate properly.

evgenux:
Thank you a lot, @stlaz! With #133 login session starts well on my pc. Even my clipmanager, which depends somehow on it, functionate properly.

evgenux:
But I noticed another related issue: when I terminate awesome wm (or qtile), a wallpaper remains and I can't get to `ly` again.

stlaz:
If it’s related only to this fix, it might be caused by https://github.com/cylgom/ly/pull/133/files#r339329964. I haven’t tested this with xserver yet, actually.

evgenux:
Yes, it is: I can't get back to `ly` only with #133. But without your patch range of programs doesn't work: for example `connman` rejects to work and prints `Error getting services` etc.

evgenux:
[...]

PR #133, opened 2019-10-25
Implement utmp audit

stlaz:
Implements utmp audit required by policykit.

This commit also flattens the pidtree for the DM by starting the
display environment directly in the first fork which already should
have the environment ready for this purpose.

Fixes https://github.com/cylgom/ly/issues/121

### DIFF
--- a/src/login.c
+++ b/src/login.c
@@ -18,6 +18,7 @@
 #include <sys/stat.h>
 #include <sys/wait.h>
 #include <unistd.h>
+#include <utmp.h>
 #include <xcb/xcb.h>
 
 int get_free_display()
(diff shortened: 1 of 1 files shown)


src/login.c
```diff
@@ -268,6 +269,38 @@ void env_xdg(const char* tty_id, const enum display_server display_server)
...
+	struct utmp *entry,
+	char *username,
+	pid_t display_pid
+) {
+	entry->ut_type = USER_PROCESS;
+	entry->ut_pid = display_pid;
+	strcpy(entry->ut_line, ttyname(STDIN_FILENO) + strlen("/dev/"));
+
+    /* only correct for ptys named /dev/tty[pqr][0-9a-z] */
+    strcpy(entry->ut_id, ttyname(STDIN_FILENO) + strlen("/dev/tty"));
```
roosemberth: Hi, could you please fix the indentation in these lines?

stlaz: oops, I thought I fixed them all. It should now be properly indented.

src/login.c
```diff
@@ -609,17 +618,28 @@ void auth(
...
+	struct utmp entry;
+	add_utmp_entry(&entry, pwd->pw_name, pid);
+
 	// wait for the session to stop
 	int status;
 	waitpid(pid, &status, 0);
+	remove_utmp_entry(&entry);
+
+	reset_terminal(pwd);
+	if (desktop->display_server[desktop->cur] == DS_XORG && xcb != NULL)
```
stlaz: I just realized this will need additional check for DS_XINITRC, I’ll fix that when I get to my laptop

stlaz: FIXED by initializing the `xcb` variable (also fixed a possible segfault), the X session still hangs as described in #121 though, gotta investigate why

cylgom:
Hi, thanks a lot! I was working on the same thing but you're far more advanced than I so we should probably merge your PR instead.

I also have the issues you and @evgenux described in #121 so I tried to investigate a little bit:
 - Xorg stays open after the desktop environment is closed
 - If you kill it, ly will hang
 - With the proprietary nvidia driver, X11 desktop environments refuse to start

Maybe my understanding is wrong, but for me:
 - We should keep that extra process to launch Xorg
 - We can't rely on the existing environment (the linux `login` from `login-utils` handles everything itself, so maybe we should, too)
 - None of this affected Sway, so the issue definitely comes from the X11 side

I am confident we will get this working though, and thanks again for contributing :)

stlaz:
Hi,
Sorry for not replying earlier, that's some good research you did.

I too noticed that Xorg stays open, I also noticed that with this patch, on logout, the pstree of the created fork gets transferred to the `init` process upon logout, which may indicate that our parent process exits, I wonder if that's caused by the terminal reset that we run or some unhandled signal sent from the child.

I think killing Xorg might be the way for now, if ly hangs, then I'll need to figure out why that is, as that should not happen.

I wonder about the proprietary driver though. I know `sway` does not like them on purpose, but I didn't know it may cause trouble to X11 desktop environments, too.

I would still like to flatten the processes if there's a way to do that.

src/login.c
```diff
@@ -526,6 +531,9 @@ void auth(
 	tb_present();
 	tb_shutdown();
 
+	// we need xcb connection to clean up after ourselves if running xorg
+	xcb_connection_t* xcb = NULL;
```
stlaz: This was silly, we will need shared memory here

stlaz:
@cylgom you were right, dealing with X11 is a bit too painful, I removed the process tree flattening for that.

Please review the changes again, see if they are ok

cylgom:
Wonderful! Thank you very much :)

PR #139, opened 2019-11-03
Mention that GNU make is required

0mp:
The current makefile is not compatible with bmake,
which is the default make on FreeBSD.


cylgom:
Could you point out what feature makes it incompatible?

0mp:
For example `$(shell ...)` and `ifeq` are are incompatible constructs. Wildcarding with `%` (e.g., `%.c`) is also not portable.

cylgom:
Thank you (:

PR #150, opened 2019-11-14
Translate Ly to Spanish

ivanhercaz:
Hi @cylgom!

I am new with Ly but due it has just a few strings to translate and it is a very beautiful display manager, I decided to translate it to Spanish.

However, the translation is pending to test because I tried to configure it and it is always in English. In `config.ini` I add `lang = es` and in then I tried to things that didn't work:

1. Translate to `es.ini` from the English version, then move it to the `/etc/ly/lang` folder.
2. Compile and install Ly again with the `es.ini` in the `/res/lang/` folder.

As I said, both didn't work. Could you help me with this issue before to merge the PR?


cylgom:
Hi! I just fixed the issue in 676cbf5 so we can merge. Thank you for your contribution (and patience) :)

PR #158, opened 2019-12-10
Remove /usr/bin/env from default path

themadprofessor:
The path variable should only contains directories, and `/usr/bin/env` is a file.
While shells will generally ignore non-directories in the path, other software doesn't. For example `rkhunter` raises an error.


cylgom:
Don't know how this happened, thanks :)

Issue #129, opened 2019-10-07
Plasma (Wayland) appears as Plasma


ericonr:
Hi there! I've just started using ly (really loved the fire background, btw), but I have found a small issue in it for my setup. I have both Plasma for Xorg and Plasma Wayland installed in my system. This means that ly shows two sessions with the same name. In this case, Plasma.

What is a reasonable solution for this? Should I just change the name on the session file?

For example, the way SDDM treats this is that it includes `(Wayland)` for every Wayland session. That way, both Sway and Plasma have that extra information. This could be a feature that ly supports as well, perhaps via some configuration setting.

If you come to any decision that involves adding something to ly, I would be interested in implementing it! Thanks for the attention.

PS: This [asciinema](https://asciinema.org/a/6jQ7fJrMg4HcVS3sXOa6dYLQK) shows this issue.

cylgom:
Hi, you are right we should probably do something about this.

I think the best way to achieve what we want is to change the name of the session while reading it from the session file. This can be done by writing another version of `config_handle_str` that appends ` (Wayland)` to the session file's string before saving it. This function is used in `desktop_crawl`, so we need to add a condition here.

However, some desktop environments already include ` (Wayland)` in their session file, so the new `config_handle_str` should read the input string and check if it doesn't end with something similar.

You can find the config handles and configuration options in `config.c`.
Feel free to ask any question if something is unclear :)

ericonr:
Wow, sorry for the delay. School has been crazy! Do you think there should be a config option to determine if we add `(Wayland)` to the session name?

ericonr:
Well, this one's done! I will close it :)

PR #162, opened 2019-12-24
Add Wayland specifier to session names

ericonr:
This PR is related to issue #129, adding a config option that makes ly add a " (Wayland)" specifier to Wayland session names, the way SDDM, for example, does. This this makes it possible to differentiate Plasma on Xorg from Plasma on Wayland, for example.

I am unsure whether I should verify anything regarding the size of the "name" string, given that I simply concatenate the strings. I had added a realloc call to make it bigger, but it seems that argoat + configator don't use any dynamic allocations, so it wouldn't be the correct way of doing this.

In the interest of future proofing, do you believe it would be worth it to add an option for a Xorg specifier as well?

### DIFF
--- a/res/config.ini
+++ b/res/config.ini
@@ -83,6 +83,9 @@
 # wayland setup command
 #wayland_cmd = /etc/ly/wsetup.sh
 
+# add wayland specifier to session names
+#wayland_specifier = false
+
 # wayland desktop environments
 #waylandsessions = /usr/share/wayland-sessions
 
--- a/src/config.c
+++ b/src/config.c
@@ -184,6 +184,7 @@ void config_load(const char *cfg_path)
 		{"term_reset_cmd", &config.term_reset_cmd, config_handle_str},
 		{"tty", &config.tty, config_handle_u8},
 		{"wayland_cmd", &config.wayland_cmd, config_handle_str},
+		{"wayland_specifier", &config.wayland_specifier, config_handle_bool},
 		{"waylandsessions", &config.waylandsessions, config_handle_str},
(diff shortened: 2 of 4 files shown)


ericonr:
`#wayland_specifier = true`

Just realized I left it as true in the config comment, but it's set to false by default in the code. Should I keep the pattern and change it to false in the config?

ericonr:
I have to take a closer look at argoat / configator, but I need a guarantee that the name string is null terminated, so searching for a substring is safe (can't segfault), and that concatenating `" (Wayland)"` doesn't make the string too big. Gotta make sure the program can't crash randomly.

ericonr:
I have added back the `realloc` call. Taking a proper look at the `config_handle_str` code showed me that it makes use of dynamic allocation for the string. I believe this could be documented on the function or close to it, perhaps.

stlaz:
Wouldn't it be easier to just fix your session file instead?

ericonr:
Most session files for Wayland don't include the specifier, and they might be overwritten by your package manager. I could check if GDM also adds the specifier, but at least SDDM does, so it's a solution that has been used. Making it optional also helps in keeping ly customizable.

cylgom:
@stlaz for us sure it would be easier, for the end-user not so much. IMHO this change is worth the extra code because it makes tinkering a lot easier (you can install tons of desktop environments without getting lost, for example)
@ericonr I'd like the config to read the default setting (false), but you can add another commented line just under the first one with the value set to true: I've been selfishly doing that to retain my favorite settings :)

ericonr:
Just fixed it, then!

cylgom:
Thanks :)

PR #163, opened 2019-12-24
Make the project compilable with clang.

ericonr:
Clang generates more warning messages, such as pointing out a lack
of newline at end of file and issues with the size of a strncopy.

Moving -Wno-unused-parameter to the end of the flags avoids it being
overwritten by -Wextra.

Removing -Werror avoids compilation errors that may arise from new
compiler warnings that might come from newer versions of compilers.

Additionally, the -Werror flag made clang error out when compiling ly before the warnings it detected were corrected.

### DIFF
--- a/makefile
+++ b/makefile
@@ -1,7 +1,7 @@
 NAME = ly
 CC = gcc
 FLAGS = -std=c99 -pedantic -g
-FLAGS+= -Wall -Wno-unused-parameter -Wextra -Werror=vla -Werror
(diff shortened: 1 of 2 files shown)


ericonr:
Out of curiosity, all `pars_count` parameters are unused for now. What's their intended purpose?

cylgom:
Thank you @ericonr;
`pars_count` gives the length of `pars`, which is a string array.
It's not used by Ly but the goal of argoat (and configator) was to use a single
function pointer format for the sake of simplicity and interoperability :)

PR #168, opened 2020-01-27
Adding desktops that are tested

RandomGuyDTB:
I personally tested the latest version (as of right now) of MaXX Interactive Desktop and the latest version of GNU WindowMaker with Ly and I can confirm they work fine.


cylgom:
Thank you!

Issue #167, opened 2020-01-26
Issue with Sway 1.4


rhysperry111:
There are issues starting Sway with Ly in it's latest version. 

This problem is caused by the DISPLAY variable being set. I have fixed it by adding `unset DISPLAY` in wsetup.sh but I am not sure how you guys would want it implemented

ivanhercaz:
Thank you for this report, @rhysperry111. I am considering to switch to Sway from i3, and it is very useful to know that.

ericonr:
Furthermore, I've found that XDG_SESSION_TYPE ends up being set as `tty` instead of `wayland`.

stlaz:
for the record, here's the debug log from sway for the failed sessions:
```
2020-01-29 08:58:44 - [sway/main.c:152] Linux standa-arch 5.4.15-arch1-1 #1 SMP PREEMPT Sun, 26 Jan 2020 09:48:50 +0000 x86_64 GNU/Linux 
2020-01-29 08:58:44 - [sway/main.c:168] Contents of /etc/lsb-release:
2020-01-29 08:58:44 - [sway/main.c:152] LSB_VERSION=1.4
2020-01-29 08:58:44 - [sway/main.c:152] DISTRIB_ID=Arch
2020-01-29 08:58:44 - [sway/main.c:152] DISTRIB_RELEASE=rolling
2020-01-29 08:58:44 - [sway/main.c:152] DISTRIB_DESCRIPTION="Arch Linux"
2020-01-29 08:58:44 - [sway/main.c:168] Contents of /etc/os-release:
2020-01-29 08:58:44 - [sway/main.c:152] NAME="Arch Linux"
2020-01-29 08:58:44 - [sway/main.c:152] PRETTY_NAME="Arch Linux"
2020-01-29 08:58:44 - [sway/main.c:152] ID=arch
2020-01-29 08:58:44 - [sway/main.c:152] BUILD_ID=rolling
2020-01-29 08:58:44 - [sway/main.c:152] ANSI_COLOR="0;36"
2020-01-29 08:58:44 - [sway/main.c:152] HOME_URL="https://www.archlinux.org/"
2020-01-29 08:58:44 - [sway/main.c:152] DOCUMENTATION_URL="https://wiki.archlinux.org/"
2020-01-29 08:58:44 - [sway/main.c:152] SUPPORT_URL="https://bbs.archlinux.org/"
[...]

PR #170, opened 2020-01-29
Unset DISPLAY variable for wayland

rhysperry111:
See #167 
This is the fix that I currently use to get ly to work with sway. I am not sure if this is the proper way to implement it as I have only tested it on bash and zsh. Also to note is that it does not fix the problem of XDG_SESSION_TYPE being set to `tty`

### DIFF
--- a/res/wsetup.sh
+++ b/res/wsetup.sh
@@ -6,6 +6,7 @@
 # Copyright (C) 2001-2005 Oswald Buddenhagen <ossi@kde.org>
 
 # Note that the respective logout scripts are not sourced.
+unset DISPLAY
 case $SHELL in
   */bash)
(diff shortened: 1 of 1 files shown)


rhysperry111:
Why are checks failing? The area highlighted as "failed" was there before I modified the code

cylgom:
@rhysperry111 Well CodeFactor is supposed to generate the quality badge you can see in the readme, but promoted itself to code-reviewing for some reason. Don't pay attention. I am taking a look at the issue with Sway right now, it could take a while.

cylgom:
As explained [here](https://github.com/cylgom/ly/issues/167#issuecomment-581062483), #167 is now fixed internally. Thanks again :)

PR #172, opened 2020-01-31
fix hidden cursor after logout

### LINKED ISSUE #118: No cursor after exit
Hey! Your DM is just awesome. Really. A simple DM in TUI... Ok, the problem is that if systemctl launches it, there's no problem, there's a cursor, I can connect me. But when I'm connected, if I exit my session (Win+Shit+E in i3 or just `exit` in shell) ly comes again and... No cursor anymore. I can't see on what line I am. I can still write or even select the DE but I can't see where I am anymore. Sorry I don't have anything to provide I really don't know what the problem could be. Of course if you need logs or something feel free to ask. I'm running Arch Linux (I installed it this morning btw so it's almost a fresh install). Thanks for answering ^^

mozartilize:
the fix is based on [Cursor disappears when running `top -n1 | head`](https://unix.stackexchange.com/questions/469770/cursor-disappears-when-running-top-n1-head)

fixes #118

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -264,6 +264,7 @@ int main(int argc, char** argv)
 				}
 
 				load(&desktop, &login);
+				system("tput cnorm");
 				break;
 			default:
 				(*input_handles[active_input])(


cylgom:
Nice one, thank you very much :)

Issue #165, opened 2020-01-20
Systemd Logind Session not closed correctly


moberrauch:
When closing sway and successfully returning to ly, the user session I just logged out from is still running. It is listed under `loginctl` and therefore also the systemd runtime directory `/run/user/1000/` is still alive / mounted.

When I relog as the same user, fortunately not a new session is opened, but the old one reused, nevertheless some issues arise from this unclean session handling, specially with dbus (e.g. controlling Auryo via `playerctl` doesn't work anymore).

**Observed behaviour:**
`loginctl` lists the user session even after exiting back to ly.

**Expected behaviour:**
The login session is closed correctly as if it never had been opened (just like after a fresh system boot).

**Reproduce:**
Log into sway from ly and exit sway again. Log in from a different tty (preferably as a different user) and execute `loginctl` and / or look for the user's runtime directory.

**System:**
Arch Linux (linux-zen 5.4.13)
Systemd version: 244.1 with PAM support
ly version: 0.4.0
sway version: 1.2

I will do further testing, if desired.

buckshee:
You should be using some type of Session Manager, a hybrid like LightDM or configuring your DE to exit in the way you want.

wolfhechel:
I think this is a bad behaviour from ly regardless if you're using a session manager or not. The problem is that the pam_start and subsequently all pam_authentication calls are done in the root ly before forking it off to xorg. This makes systemd attach the root process of ly to XDG_SESSION_ID, meaning my user session is tied to the root process.

This has multiple consequences, first of all the it's impossible to reinitialize a session after for instance killing my WM. pam_systemd will simply exit with "Cannot create session: Already running in a session or user slice". My next session will then be left without a XDG_SESSION_ID initialized.
The second consequence is that if I do loginctl terminate-session $XDG_SESSION_ID ly simply exits. In most DM's I've come across the behaviour is that I'll return to login prompt (owned by root).

I'm working on a fix for this and will reference this issue once I've come up with a fix.

massimopavoni:
> I think this is a bad behaviour from ly regardless if you're using a session manager or not. The problem is that the pam_start and subsequently all pam_authentication calls are done in the root ly before forking it off to xorg. This makes systemd attach the root process of ly to XDG_SESSION_ID, meaning my user session is tied to the root process.
> 
[...]

PR #173, opened 2020-02-01
Set XDG_SESSION_TYPE earlier

loserMcloser:
Need pam_systemd module to see proper XDG_SESSION_TYPE setting, so need to export it before pam_open_session is called.


cylgom:
Hi, @roosemberth is already working on the same issue. Actually the [freedesktop](https://www.freedesktop.org/wiki/Software/systemd/writing-display-managers) documentation indicates all `XDG` vars should be initialized before the pam session is opened. This resulted in Wayland not working in PR #137. It has to do with issue #165 as well, so there is a lot going on here...

Interestingly everything is fine under Wayland with this PR, but the variable is never set under X and tty (at least on my system). We will have to figure it out entirely before I merge :)

loserMcloser:
On Mon, Feb 3, 2020, 12:12 AM cylgom <notifications@github.com> wrote:

> Interestingly everything is fine under Wayland with this PR, but the
> variable is never set under X and tty (at least on my system). We will have
> to figure it out entirely before I merge :)
>

Weird. Sorry, I didn't think to test on X and tty before submitting, my
motivation was to get pam_systemd.so to set type=wayland properly.

Thanks for your work on this!

>

roosemberth:
I think I could inspire myself from this PR to fix #137 (Sorry for the delay). I'll hopefully work on it in the upcoming days.

PR #175, opened 2020-02-03
Update wsetup.sh

stanislavzadiraev:
the DISPLAY should be unset.

PR #176, opened 2020-02-03
The DISPLAY should be unset

stanislavzadiraev:


cylgom:
?
Same as PR #175, closing...

PR #177, opened 2020-02-04
Add Romanian translation

mihaimorariu:
Thank you for this great display manager!
I would like to add the translation for the Romanian language as well.


cylgom:
Thank you :)

PR #178, opened 2020-02-04
Add /usr/sbin to the default path

0mp:


FalyPr1nc3:
These commented lines just represent default values but actually do not configure anything.
You have to update it [in the source code](https://github.com/cylgom/ly/blob/eb496b0db580e04f7726b45fc6df292662525c99/src/config.c#L284)

cylgom:
fixed in f49c537, thanks for reporting :)

PR #180, opened 2020-02-16
Add Portuguese translation.

ericonr:
Hey there!

I'd prefer to show this to a few people before getting it into ly, so I marked it as WIP. I will change it as soon as I have feedback!


laggarcia:
I would do the following modifications:

path -> caminho
inicializar -> iniciar
login -> conectar

Token and buffer do have equivalent words in portuguese, but I would say they are not used in this context.

ericonr:
I think these are sound modifications! Thanks :)

Will add them to the branch when I get home.

ericonr:
@laggarcia 
I just fixed them! Thanks again c:

ericonr:
Removed WIP flag, can be merged.

cylgom:
thanks!

PR #183, opened 2020-02-28
Fix env and utmp for X

jkrcma:
Hello,

thanks for all of your current work. I've finally managed to find some time and help you to figure out issue I originally reported (#121). I am not a C developer at all and my knowledge regarding to POSIX and Linux libraries is limited to Google Search results :laughing: so please handle with care. I do want to open up a discussion about this topic.

After I updated my upstream from October commits two things broke for me. See linked commits for more details:

1. different setting for `LC_MESSAGES` was ignored by `locale.conf` - b775f93
2. despite seeing my session in `w` policykit still requested a password in order to e.g. suspend my laptop - af90c29

I'm using a standard Arch Linux with [i3-cinnamon](https://github.com/jbbr/i3-cinnamon/tree/master/session) session.

cc @stlaz


src/login.c
```diff
@@ -572,6 +567,16 @@ void auth(
...
 
+	// get a display
+	char tty_id[3];
+	char vt[5];
+	char display_name[4];
+
+	snprintf(tty_id, 3, "%d", config.tty);
+	snprintf(vt, 5, "vt%d", config.tty);
+	// Xorg only?
+	snprintf(display_name, 3, ":%d", get_free_display());
```
cylgom, line 578: Welcome to Unix de-facto-semi-standards. I am not aware of any restriction as to how Wayland compositors should handle display management (in fact even the wsessions folder appeared as a de-facto standard). Ly currently acts as a simple launcher when it comes to Wayland compositors, so this bit is only useful for X.org desktop environments.

jkrcma: Alright so if `DISPLAY` variable truly relates only to X then a bit of refactoring should disentangle and simplify the code

PR #192, opened 2020-03-22
Fixes segfault when environment does not set LANG

### LINKED ISSUE #191: $HOME not set after login
I have two window managers (bspwm and berry) installed, which means ly presents the following login options:

- `berry`
- `bspwm`
- `xinitrc` (which only contains `exec bspwm`)
- `shell`

I use the fish shell, which gets its config from `$XDG_CONFIG_HOME/fish` (or `$HOME/.config/fish` if `$XDG_CONFIG_HOME` is not set). When I log in using any of the above 4 options, the screen flashes briefly and then returns to ly. I was able to record the screen on my phone and catch the message, which is:

```
error: Unable to locate the config directory.
error: Please set the XDF_CONFIG_HOME or HOME environment variable before starting fish.
```

So fish is telling me that `$HOME` is not set. If I change my default shell to bash I get the exact same issue when I try to log in, except there is no message. I'm assuming that bash is also unable to start because `$HOME` is undefined.

If I log in to a TTY normally, `$HOME` is set and it takes me straight to the default shell with no issues.

AdamChristiansen:
Fixes #191 by performing a null check before calling `setenv()`. A call to `setenv()` will segfault when its second argument (the variable value) is `NULL`. On some systems, `LANG` is not defined before the user logs in, so this is fixed by only exporting `LANG` if it is already defined (i.e. not `NULL`).

This change conforms to the POSIX specification. POSIX does not require `LANG` to be initialized at the time of login, like it does for variables such as `HOME`.

### DIFF
--- a/src/login.c
+++ b/src/login.c
@@ -227,7 +227,10 @@ void env_init(struct passwd* pwd)
 	setenv("SHELL", pwd->pw_shell, 1);
 	setenv("USER", pwd->pw_name, 1);
 	setenv("LOGNAME", pwd->pw_name, 1);
-	setenv("LANG", lang, 1);
+	if (lang != NULL)
+	{
+		setenv("LANG", lang, 1);
+	}
 
 	// Set PATH if specified in the configuration
 	if (strlen(config.path))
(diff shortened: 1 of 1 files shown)


AdamChristiansen:
I didn't see #183 before I wrote this. Since this change is a subset of the features #183 adds, if #183 is merged then this can be closed.

cylgom:
I will merge #183 because it is more complete, but thank you for taking the time to investigate this.

PR #210, opened 2020-06-03
Add more details to the Debian dependencies

devenblake:
Instead of saying Debian-based distros need `libpam0g-dev` or whatever I just put an `apt` command that downloads all the dependencies (build-essential, libpam0g-dev, libxcb-xkb-dev).


cylgom:
Thank you :)

PR #251, opened 2020-11-14
Add location of Xresources for XDG Base Dir Spec

kinjalik:
Some of users want to keep their configs in unified folder. XDG Base Directory allows to do so, but there is no support of than in ly

### DIFF
--- a/res/xsetup.sh
+++ b/res/xsetup.sh
@@ -90,6 +90,7 @@ elif [ -f /etc/X11/Xresources ]; then
   xrdb -merge /etc/X11/Xresources
 fi
 [ -f $HOME/.Xresources ] && xrdb -merge $HOME/.Xresources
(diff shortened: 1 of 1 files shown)


AnErrupTion:
@kinjalik I don't understand. The line above is for ~/.Xresources, isn't that the correct path?

UtkarshVerma:
While this could be merged. Based on the code we have, we could already achieve the desired functionality.

We can specify `xnitrc` in `/etc/ly/config.ini`:
```
xinitrc = ~/.config/X11/xinitrc
```

This will tell `ly` to source xinitrc from this custom path. I merge xrdb in my xinitrc which makes everything work out.

AnErrupTion:
> This will tell ly to source xinitrc from this custom path. I merge xrdb in my xinitrc which makes everything work out.

The problem about this is that not everybody launches their DE/WM from an xinitrc file. By directly including it in the xsetup.sh, it ensures everything can be merged even if they select another entry than xinitrc.

PR #252, opened 2020-11-16
Update readme.md

plasmoduck:
add dwm to support list. Been using with dwm on multiple machines inc FreeBSD


nullgemm:
Just came out of the grave. Thanks for your contribution :)

PR #281, opened 2021-04-24
Allow en.ini load and add config to hide f1 commands

voidraen:


AnErrupTion:
Why force load the en.ini file knowing it already uses that file's content as defaults? (The file isn't read)

aethieria:
So that you can customize the values without having to rebuild the project

PR #282, opened 2021-04-26
asterisk_empty

woowgers:
Some login utilities don't write any output during the password input, so some users would (and I do) like ly to behave like them.

### DIFF
--- a/res/config.ini
+++ b/res/config.ini
@@ -9,6 +9,10 @@
 #asterisk = *
 #asterisk = o
 
+# print nor asterisk and password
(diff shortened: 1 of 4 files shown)


nullgemm:
We don't need a specific option to achieve this, simply replace the asterisk character by a space in the config

woowgers:
Okay, but I mean, space is visible by cursor movement to the right, so that
it is not too hard to guess the length of the password.

сб, 23 апр. 2022 г. в 20:20, nullgemm ***@***.***>:

PR #283, opened 2021-05-02
Matrix scrolling text animation

StaleHyena:
Thought it would look cool for a rice :P
Mods the already existing fire animation to allocate a state object,
as to keep the animation and term_buf data separate.

Doesn't leak memory with any combination of config values i
tested. I tried to follow the existing allocation error handling style,
not sure if it is sound!

Would appreciate feedback if any.


mevangelisti:
This is cool as hell, nice work.

mjevange:
any change we can pull these? o:)

PR #286, opened 2021-05-09
add qtile to list of supported de/wm

yobleck:
caveat: only tested on one machine


fiorematteo:
I successfully used ly with qtile  on my machine too

PR #290, opened 2021-05-26
fix: fix typo

xxxxxliil:


nullgemm:
I don't think this is really needed, but thanks for caring :)

PR #303, opened 2021-08-15
Check if lang is NULL and use shorter check

qontinuum-dev:
This fixes an issue that causes a problem on non-systemd systems and that is anyway a programming error.
It also uses shorter checks, I would obviously use the same as you if you prefer.

### DIFF
--- a/src/login.c
+++ b/src/login.c
@@ -212,22 +212,14 @@ void env_init(struct passwd* pwd)
 	char* lang = getenv("LANG");
 	// clean env
 	environ[0] = NULL;
-
-	if (term != NULL)
-	{
-		setenv("TERM", term, 1);
-	}
-	else
-	{
-		setenv("TERM", "linux", 1);
-	}
-
+	
+	setenv("TERM", term ? term : "linux", 1);
 	setenv("HOME", pwd->pw_dir, 1);
 	setenv("PWD", pwd->pw_dir, 1);
 	setenv("SHELL", pwd->pw_shell, 1);
 	setenv("USER", pwd->pw_name, 1);
 	setenv("LOGNAME", pwd->pw_name, 1);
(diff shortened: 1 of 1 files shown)


src/login.c
```diff
@@ -212,22 +212,14 @@ void env_init(struct passwd* pwd)
...
-
+	
+	setenv("TERM", term ? term : "linux", 1);
 	setenv("HOME", pwd->pw_dir, 1);
 	setenv("PWD", pwd->pw_dir, 1);
 	setenv("SHELL", pwd->pw_shell, 1);
 	setenv("USER", pwd->pw_name, 1);
 	setenv("LOGNAME", pwd->pw_name, 1);
-	setenv("LANG", lang, 1);
+	setenv("LANG", lang ? lang : "en_US.UTF-8, 1);
```
negativeExponent, line 222: missing closing quote "

i did the same thing for s6/obarun as workaround. though this works, this will likely break other systems that are not using EN locale since  LANG and the rest will not be set anymore through /etc/profile, or skips /etc/vconsole.conf. ly has to run after LANG is normally set by system.

qontinuum-dev: Thanks for the double quote

Concerning the breakage that may occur, I don't see what is the issue with the things you mentionned. The value is just a default and is set in the case LANG is not already set. There is no bypass of the stuff set in /etc/profile and /etc/vconsole.conf using this method, is there?

sebastka:
How about this instead? 98c65b3e9b32bc917d657406929d556fafea5c5e

nullgemm:
> How about this instead? [98c65b3](https://github.com/fairyglade/ly/commit/98c65b3e9b32bc917d657406929d556fafea5c5e)

I think this is better, can you do this @qontinuum-dev ?

PR #310, opened 2021-10-02
Add support to build from copr

dhalucario:


nullgemm:
Wow, that's pretty cool, I can't try it right now but will definitely give it a try next week :+1:

nullgemm:
So I installed Fedora 35 yesterday but couldn't make this work. I get a black screen with and without this patch. Is there anything specific that must be done with this distro (other than `sudo make installselinux`)?

PR #315, opened 2021-10-31
Add matrix animation.

carterhay:
Adding a new animation. This is a simplified animation of the terminals in the movie "The Matrix".


nullgemm:
merged #283 instead, but thanks for submitting this :)

PR #320, opened 2021-11-13
Use tb_poll_event instead of tb_peek_event

jakobrs:
Reduces CPU usage noticeably on my system, from 14% to nearly 0%. There is an issue where the cursor doesn't get positioned correctly on startup, this is because the cursor position is set in `input_handles[active_input]` on line 144 (which calls tb_set_cursor), before `position_inputs` has been called. It is not enough to just call `position_inputs` before line 144, because `position_inputs` itself requires that `draw_box` has been called,

There are other issues with this approach as well, if I can't find a solution for them it might be easier to just increase the default min_refresh_delta.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -141,10 +141,16 @@ int main(int argc, char** argv)
 	struct term_buf buf;
 	uint8_t active_input = config.default_input;
 
-	(*input_handles[active_input])(input_structs[active_input], NULL);
 
 	// init drawing stuff
 	draw_init(&buf);
+
+	// draw_box and position_input are called because they need to be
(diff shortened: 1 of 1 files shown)


nullgemm:
This is definitely a step in the right direction, however it breaks animations: we should decide which method to use depending on the current configuration. If the animations are enabled, we should peek events. If not, then we can poll. If you can make these changes I will merge this asap :)

jakobrs:
I added a call to `*input_handles[active_input]` in the `if (update)` block, as this seems to fix the cursor positioning issues.

nullgemm:
thanks a lot!

PR #330, opened 2022-01-02
Use XDG_RUNTIME_DIR for storing Xauthority

UtkarshVerma:
This PR saves `lxauth` in `XDG_RUNTIME_DIR`.

### DIFF
--- a/src/login.c
+++ b/src/login.c
@@ -304,10 +304,10 @@ void remove_utmp_entry(struct utmp *entry) {
 	endutent();
 }
 
-void xauth(const char* display_name, const char* shell, const char* dir)
+void xauth(const char* display_name, const char* shell)
 {
 	char xauthority[256];
-	snprintf(xauthority, 256, "%s/%s", dir, ".lyxauth");
+	snprintf(xauthority, 256, "%s/%s", getenv("XDG_RUNTIME_DIR"), "lyxauth");
 	setenv("XAUTHORITY", xauthority, 1);
 	setenv("DISPLAY", display_name, 1);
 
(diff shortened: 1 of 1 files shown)


src/login.c
```diff
@@ -304,10 +304,10 @@ void remove_utmp_entry(struct utmp *entry) {
 	endutent();
 }
 
-void xauth(const char* display_name, const char* shell, const char* dir)
+void xauth(const char* display_name, const char* shell)
 {
 	char xauthority[256];
-	snprintf(xauthority, 256, "%s/%s", dir, ".lyxauth");
+	snprintf(xauthority, 256, "%s/%s", getenv("XDG_RUNTIME_DIR"), "lyxauth");
```
nullgemm, line 310: We have to test whether this var is empty or not! If it is, we could try using `XDG_CONFIG_HOME` instead, and if it also is empty, we can use `pwd->pw_dir` as a last resort.

src/login.c
```diff
@@ -343,18 +343,10 @@ void xorg(
 	const char* vt,
 	const char* desktop_cmd)
 {
-	// generate xauthority file
-	const char* xauth_dir = getenv("XDG_CONFIG_HOME");
-
-	if ((xauth_dir == NULL) || (*xauth_dir == '\0'))
-	{
-		xauth_dir = pwd->pw_dir;
-	}
```
nullgemm, line 352: This is the kind of test we should be using

nullgemm (commented):
Cool, I have made a few comment, once everything is ready this is definitely something I can merge :)

UtkarshVerma:
Implemented your suggestions in the last commit. Please have a look.

PR #352, opened 2022-05-12
Changed active_input selection to be based on default config

kefoster951:
changed how the curser placement works to be based on if there is a saved username or not. if there is a saved user, it started on the password field, if there isn't a user name it will start on the username field.

### DIFF
--- a/src/config.c
+++ b/src/config.c
@@ -271,7 +271,7 @@ void config_defaults()
 	config.blank_box = true;
 	config.blank_password = false;
 	config.console_dev = strdup("/dev/console");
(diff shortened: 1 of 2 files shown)


AnErrupTion:
@kefoster951 I think a better way is to first modify the config's default input to be on the username field, and is there is already a saved username, then it should automatically be on the password field.

kefoster951:
@AnErrupTion  I think you are right, I updated the default setting to be the user field, and if there is a user saved it goes to the password field automatically.

if you set the config option to with the session_swtich or the password_input, it will put the curser there.

PR #354, opened 2022-05-13
Make TAB key loop through U.I elements infinitely

SapphireAmoeba5:
The old is behaviour is that once the TAB key reached the password field, it would stop looping through the fields. This change allows the tab key to infinitely loop through each U.I element.


kefoster951:
@SapphireAmoeba5 you should use SESSION_SWITCH instead of 0, it's more clear what the code is doing

Issue #295, opened 2021-07-20
Feature: support for openrc


gerardet46:
Hello,
it would be great to have support for OpenRC

Cavernosa:
Can you try adding `-nl /usr/bin/ly` to agetty_options in `/etc/conf.d/agetty.tty1` and see if it works fine?

Issue #216, opened 2020-07-01
Login failure and a new release


cinerea0:
First, I wanted to report that I'm experiencing the same problems listed in #36 : Attempting to log on to any profile (shell, .xinitrc, Plasma), simply causes ly to display "Logged out" in the box. The closest thing to a log of this I can find is the following (note the timestamps):

```
2020-07-01T15:06:44.20217 authpriv.info: Jul  1 11:06:44 ly: pam_unix(login:session): session opened for user erebus by LOGIN(uid=0)
2020-07-01T15:06:44.24093 authpriv.info: Jul  1 11:06:44 ly: pam_unix(login:session): session closed for user erebus
```

As an aside, I think the improvements made in #189 are significant enough to warrant at least a new minor release, since they fix a lot of PAM issues.

cinerea0:
Wanted to give an update, this same problem occurs on the 0.5.2 tagged release:

```
2020-08-01T14:26:40.11819 authpriv.info: Aug  1 10:26:40 ly: pam_unix(ly:session): session opened for user erebus by LOGIN(uid=0)
2020-08-01T14:26:40.17513 authpriv.info: Aug  1 10:26:40 ly: pam_unix(ly:session): session closed for user erebus
```

ericonr:
This is caused by `env_init` in `src/login.c`, https://github.com/nullgemm/ly/blob/master/src/login.c#L208-L242

`LANG` isn't guaranteed to be set, so it segfaults when trying to access `lang`, which in our case was set to `NULL`. That said, I'm not sure I see the purpose of the `setenv("TERM", term, 1);` and `setenv("LANG", lang, 1);` lines.

negativeExponent:
> This is caused by `env_init` in `src/login.c`, https://github.com/nullgemm/ly/blob/master/src/login.c#L208-L242
> 
> `LANG` isn't guaranteed to be set, so it segfaults when trying to access `lang`, which in our case was set to `NULL`. ~That said, I'm not sure I see the purpose of the `setenv("TERM", term, 1);` and `setenv("LANG", lang, 1);` lines.~ I see it resets the environment.

wow thanks for this tip. In my case (Obarun S6/66), LANG is null during init so hence i get login problems (immediately shows logged out once username/pass has been entered regardless of session to start).

SInce **TERM** has an override in **env_init**, might as well add a fallback for **LANG** which worked in my case:

https://gist.github.com/negativeExponent/a7c3181a2c736fecf76410989f412adc

Issue #207, opened 2020-04-29
Background Color


skyexton:
How can I change the background color?

EgorDuplensky:
**Configuration**
You can find all the configuration in /etc/ly/config.ini. The file is commented, and includes the default values.

flatcap-monty:
> Can I Use Hex Values?
I would also like to know. The configuration is commented, but I'm not certain what the numbers for the background colour correspond to?

xenbeta:
Same Question

Kharacternyk:
You may find this link helpful: https://askubuntu.com/questions/147462/how-can-i-change-the-tty-colors

takagiy:
@xenbeta 
I noted a workaround which works for me to use hex colors in #146.

Neules:
Well, to be honest `config.ini` isn't so much commented. The text `background color id` is quite ambiguous. It could, at least, include a range of id's to be used.

In my case, changing this value does nothing. My screen keeps black/white.

Eklektisk:
You can find a list of color codes in [termbox_next/termbox.h](https://github.com/cylgom/termbox_next/blob/master/src/termbox.h):
```
// Colors (see struct tb_cell's fg and bg fields).
#define TB_DEFAULT 0x00
#define TB_BLACK   0x01
#define TB_RED     0x02
#define TB_GREEN   0x03
#define TB_YELLOW  0x04
#define TB_BLUE    0x05
#define TB_MAGENTA 0x06
#define TB_CYAN    0x07
#define TB_WHITE   0x08
```
`TB_DEFAULT` is analogous to `TB_BLACK` for background colors and `TB_WHITE` for foreground colors. All other termbox colors are equivalent to their system color codes (`TB_BLACK` represents color0, `TB_RED` represents color1, etc...)

As a side note, the `bg` option only affects the background color for the login box and the shutdown/reboot options.

Neules:
Seriously, you should do the thing less tricky. Why not use normal strings or integers as normal people, like "red", "blue" or 2, 3, 4. I'm afraid not everyone understands this "colour table code" (not even me, I'm not a programmer!).

plasmoduck:
How did you get the whole background that dark blue/greyish color?

TechieAndroid:
I'm also curious as to how you changed the background color entirely.

Figuera:
I had this same problem and suffered a little to solve it. I will post here a summary of my research as a reference for the future:

[...]

PR #368, opened 2022-05-17
OpenRC service and support for more inits

### LINKED ISSUE #343: [OpenRC] Can't shutdown or reboot after login
I'm using the OpenRC script from the [ly-gentoo fork](https://github.com/pspiagicw/ly-gentoo) with LANG patch from [here](https://gist.github.com/negativeExponent/a7c3181a2c736fecf76410989f412adc).

After login it's impossible to shutdown or reboot with `loginctl poweroff` command or any other poweroff command. The only way is pushing the power button, or logging out and pressing F1 or F2.

Cavernosa:
I tested and it doesn't seem to break anything

Closes #343 #295


wncry:
i am still trying to get this branch working on gentoo.

`start()` in `res/ly-openrc` does not execute.
i have been playing around with other commands to try to
fix this, and i will let you know when i have solved the issue.

Cavernosa:
Weird.. if it's running openrc it should run fine since it's the exact same init and afaik it's the same structure.
You could try changing the shebang to /usr/bin/openrc-run or something, or using getty instead of agetty. (if you haven't tried these already)
That, or it's just spawning in another tty that you're not looking xd. That's all i can think of...

gagero:
@wncry Did you manage to get it working?

Cavernosa:
>Also I don't know if this a case specific issue (I am testing on artix linux) but I will start on tty2, no matter what the script does. So it might just be best to use tty2 (It might also be useful to note that the systemd service file uses tty2).

The script only starts ly in the provided tty, it's ly that's teleporting you to tty2, try to edit the `tty` in `config.ini` to be the same as the init. (sorry, i guess i forgot to say that)
I'll try to come with something so that the two values sync automatically.

I'm using tty7 because i thought it's easier, for the 1-6 ones you have to disable the `agetty.tty*` service so that the normal agetty doesn't override the ly agetty (assuming you have them enabled by default)

> Instead it's recommended to use [supervisor](https://wiki.gentoo.org/wiki/OpenRC/supervise-daemon#General_recipe).

Yeah, that's fine i guess.

> I made a fork with my version of the [script](https://github.com/MadcowOG/ly/blob/f05252ec77e98fef1dc7c9bcb9fd0464ab5e0570/res/ly-openrc) feel free to just steal it for this PR. I did use [this script](https://github.com/pspiagicw/ly-gentoo/blob/89a580f65fea8a35a1df64ecb92367a327c2bb79/res/ly) from [pspiagicw](https://github.com/pspiagicw) to start out because it worked somewhat.

pspiagicw's script clogs up openrc-run which introduces a lot of issues, read my issue here where i <s>tried to</s> explain: #343
Basically you shouldn't be using `exec ly` directly because it doesn't fork the process, thus clogging up openrc-run.

MadcowOG:
I see now! I'm thinking this should be mentioned in the documentation. Although this could be intuited, some users may not intuit this.
I changed the script to fit these needs, would you say it works for the use case?

Cavernosa:
> I see now! I'm thinking this should be mentioned in the documentation. Although this could be intuited, some users may not. 
I changed the [script](https://github.com/MadcowOG/ly/blob/openrc/res/ly-openrc) to fit these needs, would you say it works for the use case? It is essentially the same as your script, but it follows the guidelines in the [OpenRC Script Guide](https://github.com/OpenRC/openrc/blob/master/service-script-guide.md#dont-write-your-own-startstop-functions) a bit more closely. 

Yeah. Nice script, i'll try it out but i'm pretty sure it works.
One thing though, i think the TTYCONF can be boiled down to just 

`cat /etc/ly/config.ini | sed -n 's/^tty.*=[^1-9]*// p'`

> ~For automatically syncing with config.ini do you think that using sed would be appropriate?~ I got the script to pull from the config for tty. I have tested it and it appears to work.

lmao that's exactly what i was doing

MadcowOG:
Sounds great! It might be useful to mention that the method mentioned in [#207](https://github.com/fairyglade/ly/issues/207#issuecomment-716212067) by [Figuera](https://github.com/Figuera); for changing the background and foreground color is still possible. By adding this line (Best to place it after either CONFTTY or TTY and depending on which var; the var in place of the tty will need to be changed, of course):
``` Bash
/usr/bin/printf "%b" "\e]P0{background-color}\e]P7{foreground-color}\ec" > /dev/"${TTY}"
```

MadcowOG:
I'm recommending that along with this PR there should be a version bump so that the non-git packages also come with the fix to the issue talked about in #216 which prevents logging in, at least with systems not running systemd.

Cavernosa:
The background color changer could be added but i don't think the init script is the best place to put that. It should be a built-in option in ly .
Also @MadcowOG i think you forgot to accept my [PR](https://github.com/MadcowOG/ly/pull/1) so i can merge the changes

MadcowOG:
My apologies I didn't see that. I have merged it.

MadcowOG:
> The background color changer could be added but i don't think the init script is the best place to put that. It should be a built-in option in ly .

I would like the same thing but nullgemm has said that he doesn't plan on supporting that feature, [#207](https://github.com/fairyglade/ly/issues/207#issuecomment-1110185757). So it's more of a workaround.

MadcowOG:
One thing that is a bit annoying that could be a good change is to move the installation of each service file to their own makefile functions (like with openrc atm). Because install just always installs the systemd service file even if you are on openrc. It might require a bit of work downstream but still might be worth it for compatibility.

MadcowOG:
I made a [new branch](https://github.com/MadcowOG/ly/tree/make-services) which contains a modified [makefile](https://github.com/MadcowOG/ly/blob/make-services/makefile) and [readme](https://github.com/MadcowOG/ly/blob/make-services/readme.md). I think this could provide better transparency over different systems.

MadcowOG:
@wncry Would you mind testing the latest script in this PR to confirm that it works on Gentoo?

AnErrupTion:
@MadcowOG @Cavernosa Can this be merged?

Cavernosa:
I think so. I don't have any new ideas and the script is working fine.

AnErrupTion:
Alright then!

PR #379, opened 2022-05-22
add instruction of aur package installation to readme

KR1470R:
I've published [ly on AUR ](https://aur.archlinux.org/packages/ly-aur)


wncry:
https://aur.archlinux.org/packages/ly-aur
returns 404

i think it can be found here??
https://aur.archlinux.org/packages/ly
idk

Cavernosa:
Isn't it better to link to the AUR instead of yay? since not everyone uses yay

PR #409, opened 2022-06-20
zsh environment fix

### LINKED ISSUE #407: Not setting ZSH environment correctly
I'm using Ubuntu 21.10 (Note with Regolith mashed on top of GNOME). Was using default gdm3.

When logged in via gdm3 my PATH is set as
```
/home/shiny/.local/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games:/snap/bin:/snap/bin:/usr/local/go/bin:/home/shiny/.poetry/bin
```

When logged in through Ly:
```
/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games:/snap/bin
```

This is missing the 2 `/home/shiny/` dirs (first and last) as well as `/usr/local/go/bin`

Possibly related issues: #357, #405

Shiny380:
Resolves #407 

Since ZSH is a Bourne compatible shell, it should source both `/etc/profile` and `$HOME/.profile`

### DIFF
--- a/res/wsetup.sh
+++ b/res/wsetup.sh
@@ -24,6 +24,8 @@ case $SHELL in
     [ -d /etc/zsh ] && zdir=/etc/zsh || zdir=/etc
     zhome=${ZDOTDIR:-$HOME}
     # zshenv is always sourced automatically.
+    [ -f /etc/profile ] && . /etc/profile
+    [ -f $HOME/.profile ] && . $HOME/.profile
     [ -f $zdir/zprofile ] && . $zdir/zprofile
     [ -f $zhome/.zprofile ] && . $zhome/.zprofile
     [ -f $zdir/zlogin ] && . $zdir/zlogin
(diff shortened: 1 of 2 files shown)


mkaito:
This is perhaps a distro-specific matter, but on my system (Arch), `/etc/zsh/zprofile` contains the following:

```
❯ cat /etc/zsh/zprofile
emulate sh -c 'source /etc/profile'
```

Which should in theory take care of the matter without explicitly sourcing `/etc/profile` again.

negativeExponent:
when using zsh, you are suppose to normally use .zprofile (and other similar or equivalent). just symlink .profile to .zprofile if you only specifically want to use zsh instead of bash.

there is no need for this PR honestly

Issue #346, opened 2022-04-27
settings problem


ArchKubi:
there is no document and wiki
ı search change color but here is no beginner solution
and more important there is no man page ???? why
![2022-04-27-211225_1280x1012_scrot](https://user-images.githubusercontent.com/80429360/165592162-03cd9ab6-3b48-4c64-b166-943997d822a0.png)
thanks for this project but we need document for beginner user

AnErrupTion:
I think what we need to do is in the help command of Ly, instead of telling the user to RTFM (which doesn't exist), we should tell something like "Please modify config values at /etc/ly/config.ini".

Issue #260, opened 2020-12-17
Question: Color scheme


rubenkharel:
Which file is responsible for the colors?
I want to modify it a bit with gruvbox scheme.

rubenkharel:
Nvm, I found it, there is `config.ini.` inside `res` directory.

ThatOneCalculator:
Wait, how did you do it? In `config.ini`, I see 
```
# background color id
#bg = 0
```, but no way to set hex codes or anything.

ArchKubi:
> Nvm, I found it, there is `config.ini.` inside `res` directory.

![2022-04-27-210735_1280x1012_scrot](https://user-images.githubusercontent.com/80429360/165591393-a3c2bfd3-6afb-495c-a31a-8adabee12ad5.png)
are you sure ?

Issue #238, opened 2020-09-21
ly wont accept background color id


plasmoduck:
I've tried changing 0 to 243 4 etc also html colors like #282828 but it just stays black.

schijtreiger:
0x2434 does not work?

Issue #146, opened 2019-11-06
Background is white and can't be changed


maxigaz:
I've tried both ly 0.4.0 and the latest git version (installed from the AUR packages `ly` and `ly-git`, respectively) under Manjaro Linux + i3 and Sway. In both cases, the background is almost always white, although the line `bg=0` is commented out in `/etc/ly/config.ini` (which should default to black, I assume?). This makes the text somewhat hard to read.

I’ve tried uncommenting that line and I get the same results.

Also, I’ve found an example of setting the background and foreground in #80, but from my understanding, it is mentioned there that this option wouldn't work under tty. I don’t know if the situation has changed since then and if it has, what format/value is accepted in the current version.

takagiy:
I hit the similar situation; In my case the bg and the fg were always black and white respectively and config did not seem to affect. I don't know what I'm missing, but I found a workaround finally.
I edited the `/lib/systemd/system/ly.service` appending a `ExecStartPre` line to modify tty's color pallet.

```
[Unit]
Description=TUI display manager
After=systemd-user-sessions.service plymouth-quit-wait.service
After=getty@tty2.service

[Service]
Type=idle
ExecStartPre=/usr/bin/printf '%%b' '\e]P0FFC5C4\e]P7FFFFFF\ec'
ExecStart=/usr/bin/ly
StandardInput=tty
TTYPath=/dev/tty2
TTYReset=yes
TTYVHangup=yes

[Install]
Alias=display-manager.service
```

The above example sets color0(bg) to `#FFC5C4` and color7(fg) to `#FFFFFF`.

takagiy:
I found that `term_reset_cmd` in `/etc/ly/config.ini` also has to be modified to keep the colors after logging-in and while in second time login screen after logging-out again.

```
term_reset_cmd = /usr/bin/tput reset; /usr/bin/printf '%b' '\e]P7F1E9C7\e]P0090602\ec'
```

hiukky:
@takagiy I tried the steps of your answer but I didn’t succeed here :/

ArchKubi:
> I found that `term_reset_cmd` in `/etc/ly/config.ini` also has to be modified to keep the colors after logging-in and while in second time login screen after logging-out again.
> 
> ```
> term_reset_cmd = /usr/bin/tput reset; /usr/bin/printf '%b' '\e]P7F1E9C7\e]P0090602\ec'
> ```

P7F1E9C7 <----- is this color code?

PR #414, opened 2022-07-23
Improve config documentation

stevensonmt:
update readme to provide some direction on customization options.


AnErrupTion:
I don't really understand this. Why only document bg and fg in that case? It seems a bit useless to me.

stevensonmt:
The number of issues in the issue tracker about this very question suggested it to me. I was trying to avoid putting too much into the readme but Wilks be happy to write up documentation for other settings.

AnErrupTion:
> Your comment is unnecessarily harsh.
I didn't want to be harsh, I just expressed my confusion.

> The number of issues in the issue tracker about this very question suggested it to me.
Could you please point these out?

> If you prefer, this information could be moved to the comments in config.ini but that might get unwieldy.
I believe this is the best solution. I personally don't think this will cause much readability problems or such.

> As for only documenting fg and bg, are there other color settings available?
No, but what I meant is why you only documented the color settings. Now with your comment I do understand.

stevensonmt:
> > Your comment is unnecessarily harsh.
> 
> I didn't want to be harsh, I just expressed my confusion.

No worries. The word 'useless' seemed harsh.

> > The number of issues in the issue tracker about this very question suggested it to me.
> 
> Could you please point these out?

https://github.com/fairyglade/ly/issues/346
https://github.com/fairyglade/ly/issues/260
https://github.com/fairyglade/ly/issues/238
https://github.com/fairyglade/ly/issues/207
https://github.com/fairyglade/ly/issues/146
https://github.com/fairyglade/ly/issues/80

> > If you prefer, this information could be moved to the comments in config.ini but that might get unwieldy.
> 
> I believe this is the best solution. I personally don't think this will cause much readability problems or such.

Moved info to config.ini as comment

PR #419, opened 2022-08-08
Fixes for the XDG Base Dirs implementation (#330)

Cavernosa:
Fixes errors/typos from #330 in order to better comply with the [XDG Base Directory Specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html):
* Renamed `XDG_CONFIG_DIR` to `XDG_CONFIG_HOME`
* added `$HOME/.config/` as fallback in case XDG_CONFIG_HOME is null, and PWD as last resort.
* Filename for the PWD fallback now has a dot (`.lyxauth` instead of `lyxauth`) just like the other dotfiles in $HOME

I made some tests and it works fine. I'm a complete noob in C so I'd appreciate if someone reviews the code.

Note we should probably not use XDG_CONFIG_HOME at all since `lyxauth` is far from a config file. Using /run/user/$UID/ or /tmp/ would probably be better, and then maybe use pwd as the *very* last resort just in case, i don't really know.

### DIFF
--- a/src/login.c
+++ b/src/login.c
@@ -302,7 +302,7 @@ void remove_utmp_entry(struct utmp *entry) {
 
 void xauth(const char* display_name, const char* shell, char* pwd)
 {
-	const char* xauth_file = ".lyxauth";
+	const char* xauth_file = "lyxauth";
 	char* xauth_dir = getenv("XDG_RUNTIME_DIR");
 	if ((xauth_dir == NULL) || (*xauth_dir == '\0'))
 	{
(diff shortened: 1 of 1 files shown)


AnErrupTion (commented):
Hmm, why storing ``.lyxauth`` in the root of ``.config``? I believe having a separate directory for storing ly stuff is better than leaving it in the root of .config.

Cavernosa:
I didn't think it was necessary since it's only 1 file. But yeah having it inside a directory doesn't hurt.
As i said we could remove `XDG_CONFIG_HOME` and `~/.config` altogether but i'm not sure what would replace it... `/tmp/ly/` is my biggest bet.

AnErrupTion:
I prefer the second one (with .config/ly), I don't really like having the file in /tmp.

Cavernosa:
Fixed.

randomllama:
Why the 777 chmod?

Cavernosa:
I think i forgot to change it lol. Weirdly, in my system the ly folder always appeared as 0755 instead of 0777.

PR #431, opened 2022-08-21
Make it easy to install ly on runit systems e.g. void

erkannt:
Copied nearly wholesale from https://github.com/drozdowsky/ly-void.

I only updated the Makefile and README to match what I could see for openrc.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -111,6 +111,25 @@ If you choose a tty that already has a login/getty running (has a basic login pr
 # rc-update del agetty.tty2
 ```
 
+### runit
+
+```sh
+# make
+# make install installrunit
+# ln -s /etc/sv/ly /var/service/
+```
+
+Disable your existing display manager service if need be e.g.:
+```
+# rm /var/service/lxdm
+```
+
+If you are running on `tty2` (check your `/etc/ly/config.ini`) you can disable e.g. `agetty` running there:
+
+```
+# rm /var/service/agetty-tty2
+```
+
 ## Arch Linux Installation
 You can install ly from the [AUR](https://aur.archlinux.org/packages/ly), using yay for example:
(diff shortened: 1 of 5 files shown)


readme.md
```diff
@@ -111,6 +111,25 @@ If you choose a tty that already has a login/getty running (has a basic login pr
 # rc-update del agetty.tty2
 ```
 
+### runit
+
+```sh
+# make
```
Cavernosa, line 117: Why is there sh highlighting here? 
also `make` alone doesn't need root permissions

erkannt: I was trying to follow what came above. `sh` highlighting just out of habit. Happy to remove it.

Would you prefer: 

```
make
make install installrunit
ln -s /etc/sv/ly /var/service/
```

or:

```
$ make
# make install installrunit
# ln -s /etc/sv/ly /var/service/
```

or:

```
make
sudo make install installrunit
sudo ln -s /etc/sv/ly /var/service/
```

I've got no preferences.

AnErrupTion: Highlighting makes sense here, as other stuff above also has it.

AnErrupTion, line 117: ``make`` itself doesn't require root, so please update it to the correct symbol ($)

AnErrupTion (changes requested):

readme.md
Cavernosa: > Would you prefer:

2nd one is what i meant. Just like AnErrupTion said in the change request below.

>Highlighting makes sense here, as other stuff above also has it.

What do you mean? No other code block has `sh` highlighting and `sh` highlight is not meant for CLI (as evident here, it's thinking the commands are comments).

erkannt: Done.

PR #436, opened 2022-08-30
Fedora dependencies

jagjordi:


AnErrupTion:
@jagjordi Just a quick review- you forgot the ``dnf install`` command right at the beginning. Also, a new line would be nice, and capitalization on the Fedora name would be a nice touch as well.

jagjordi:
Addressed @AnErrupTion comments

PR #444, opened 2022-09-16
Allow arbitrary length uids when setting XDG_RUNTIME_DIR

thomasfsteeples:
When setting the XDG family of environment variables, `XDG_RUNTIME_DIR` has a hardcoded limit of 15 chars (including the null byte). However, if a user's UID is five or more digits long  (which is possible, and does happen in practice if you use `systemd-homed`, for instance), then this will crop the path unintentionally.

The proposed pull request fixes this by dynamically allocating a string of the correct length by counting the number of digits in a user's UID.

### DIFF
--- a/src/login.c
+++ b/src/login.c
@@ -257,16 +257,32 @@ void env_xdg_session(const enum display_server display_server)
 	}
 }
 
+static unsigned int ilog10(unsigned int n) {
+        unsigned int res = 0;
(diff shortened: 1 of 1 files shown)


AnErrupTion:
I propose instead that the max length is increased, to accomodate for longer UIDs. I don't think having a dynamic string is a good idea with such low lengths.

Issue #445, opened 2022-09-19
ly immediately displays "logged out" after entering valid credentials to log into sway


AnyTimeTraveler:
## Problem

Ly immediately displays "logged out" after entering valid credentials to log into sway.

OS: Up to date ArchLinux as of this post
Ly Version: 0.5.3.r82.g5db09ce-1 from [AUR](https://aur.archlinux.org/packages/ly-git)

## Steps to reproduce
1. Boot
2. ly appears on tty2 with sway and my username prefilled from previous logins
3. Enter Password into ly and hit enter
4. Message at the top changes to "logged out"

## Service Status

Examining the ly.service in systemd shows the following:

```
% sudo systemctl status ly.service
● ly.service - TUI display manager
     Loaded: loaded (/usr/lib/systemd/system/ly.service; enabled; preset: disabled)
     Active: active (running) since Mon 2022-09-19 22:56:42 CEST; 16min ago
   Main PID: 605 (ly)
      Tasks: 1 (limit: 38408)
     Memory: 10.6M
        CPU: 51ms
     CGroup: /system.slice/ly.service
             └─605 /usr/bin/ly

Sep 19 22:56:42 desktop systemd[1]: Started TUI display manager.
Sep 19 22:56:53 desktop ly[605]: pam_systemd_home(ly:auth): systemd-homed is not available: Unit dbus-org.freedesktop.home1.service not found.
Sep 19 22:56:53 desktop ly[605]: pam_unix(ly:session): session opened for user simon(uid=1000) by (uid=0)
Sep 19 22:56:53 desktop ly[605]: pam_systemd(ly:session): Failed to create session: Connection reset by peer
Sep 19 22:56:53 desktop ly[605]: pam_env(ly:session): deprecated reading of user environment enabled
Sep 19 22:56:53 desktop ly[605]: pam_unix(ly:session): session closed for user simon
```

## Current Workaround
1. Boot
2. ALT + CTRL + F3
3. login to tty
4. type "sway" and hit enter
5. Sway starts as it used to with ly

## Conclusion

I suspect that the line "pam_systemd(ly:session): Failed to create session: Connection reset by peer" has something to do with it,
but either my google-fu isn't strong enough or this issue is new, because I can't find anything useful.

I would appreciate suggestions on what to try very much!

the-technat:
I encountered the same issue some days ago, althouth slighlty different.

Sway logged out immediately when I wanted to login to sway. Launching the shell worked. 

[...]

PR #446, opened 2022-09-28
Re-add missing environment variables

### LINKED ISSUE #443: ly starts login-sessions as root instead of as the user who logged in
Recent versions of ly (after 86d0353) start login-sessions as root instead of as the user who logged in.

I noticed this issue, because it broke rootless Xorg, which I initially worked around by adding `needs_root_rights=yes` to `/etc/X11/Xwrapper.config`, and it also prevents a `--user` session of systemd from starting, etc.

One way to validate this is by logging in via ly and then running `loginctl list-sessions` which should show a session for the logged in user (when working), but shows a session for "root" instead (when bugged).

I've since reverted to v0.5.3 which avoids this issue and I've reverted any workarounds I had applied in the interim.

### Environment:

**Distribution**: Ubuntu 22.04
**Window manager**: [bspwm](https://github.com/baskerville/bspwm)
**Dotfiles**: [andornaut/dotfiles](https://github.com/andornaut/dotfiles)

I [install ly using Ansible](https://github.com/andornaut/ansible-workstation/blob/master/roles/desktop/tasks/ly.yml).

---

**Edit:** I've since bisected the commit history:

fairyglade/ly@86d035346e97878bd100f1e286479f4922370aa0 is the last commit that works for me.

fairyglade/ly@09ee6293bdc38d40c210d6f0bd5f811f99329086 is the first commit that exhibits the above issue.

AnyTimeTraveler:
Fixes #443 and #445 

The function `env_init` in line 622 clears the whole environment.
After adding the environment variables generated in lines 508 and 509 back in, everything works as it should.

I don't know enough about the rest of the code to know if the lines 508 and 509 can now be removed, so I didn't remove them yet.
I'd like some help from someone more experienced, if the variables are needed anywhere between the first setting (line 508) and the clearing (line 622).

You can see in #425 ([link to the specific line](https://github.com/fairyglade/ly/commit/09ee6293bdc38d40c210d6f0bd5f811f99329086#diff-07deafe51cfe4487a3ef6dcd084f1087335daa693bca590cf06653ec05618bf8L616))that the function `env_xdg` used to be called before `env_init`.

### DIFF
--- a/src/login.c
+++ b/src/login.c
@@ -618,8 +618,11 @@ void auth(
 		char vt[5];
 		snprintf(vt, 5, "vt%d", config.tty);
 
-		// set env
+		// set env (this clears the environment)
 		env_init(pwd);
+		// Re-add XDG environment variables from lines 508,509
+		env_xdg_session(desktop->display_server[desktop->cur]);
+		env_xdg(tty_id, desktop->list_simple[desktop->cur]);
 
 		if (dgn_catch())
 		{


AnyTimeTraveler:
I just booted my PC a second time and had the same problem again, but this time sway is saying that there is no seat available.
I am going to investigate further before this should be merged.

AnyTimeTraveler:
Okay, there must be some kind of race-condition here, as I have rebooted my PC 5 more times and each time logging into sway worked flawlessly.

PR #447, opened 2022-09-29
Tito integration

t0fik:
Integration with `tito` for rpm based OSes with Fedora COPR in mind.
This is different approach to #310 PR. Also COPR is capable to auto build packages using `tito`.


t0fik:
I'll create new PR with clean branch.

PR #458, opened 2022-11-12
Tidy up runit service scripts

brunorro:
* Fix the script installation path (install everything only on /etc/sv/ly)

* Update the runit scripts to get the TTY value from the /etc/ly/config.ini file

* Update readme.md

### DIFF
--- a/makefile
+++ b/makefile
@@ -100,7 +100,6 @@ installopenrc:
 
 installrunit:
 	@echo "installing runit service"
-	@install -dZ ${DESTDIR}/etc/sv/ly-runit-service
 	@install -DZ $(RESD)/ly-runit-service/* -t ${DESTDIR}/etc/sv/ly
 
 uninstall:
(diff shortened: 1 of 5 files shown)


brunorro:
Hi!! I installed ly from source in a musl void linux and I found that the runit scripts were always running on tty2

I thought it was better to get the tty value from the /etc/ly/config.init value (/bin/cat and /bin/awk will be used) and I just added a bit of info to the readme.md.

BTW, I love this project. In a musl void linux on a dell latitude d430 with 2GB of RAM is using less than 2MB of resident memory. I'm writing this from that computer (using surf as web borwser). Thank you all! :)

brunorro:
Hello again,

Just fixed the awk regex problem with good old sed backreferencing. I'm supposing anyways that cat and sed are under /bin directory (I hope it's pretty standard, though)
About those variables, I just removed them. I can't see them used and probably were copy-pasted from somewhere else

brunorro:
Sorry, had to fix the documentation: some typos in readme.md lead to format errors

PR #487, opened 2023-03-05
Change default keybindings F1/F2 to F12/F11.

lolicon0930:
Just don't want an unexpected reboot when switching back to tty2.
Maybe you can consider change the default keybind.


RannyArcher:
i like that! i wish it be merged so soon.

AnErrupTion:
@lolicon0930 Hey! While we appreciate your contribution, I personally have a few issues with it:

- You seem to have modified more files than necessary. It seems like those modifications are due to removing blank white spaces. While it's not that big of a deal, I still wonder why you did this exactly in this PR. (Unless it was automatic?)

- I don't understand your issue of having an unexpected reboot when switching to TTY2. To switch to another TTY you have to press Ctrl + Alt + F2, not just F2, so I don't see the problem here.

- Finally, to me it doesn't make sense to have F12 as shutdown and F11 as reboot, shouldn't it rather be the other way around? It would be more logical, at least in my opinion.

lolicon0930:
@AnErrupTion, hope this is some reasonable explanation:
- Yes, my editor settings include auto-removal of trailing whitespace, etc.
- I don't need `Ctrl` when not in an X session, I just need `Alt`+`F3` to switch from Ly's tty to tty3, and `Alt`+`F2` to switch back to Ly's tty. The problem is, when I hold `Alt` and alternately press the `function keys`, there is a chance to reboot. I know `Alt`+`Left/Right` can also switch between ttys, but I don't usually use it that way.
- For people like me who seldom shut down or use a computer as a server, F11 as shutdown and F12 as reboot are indeed more reasonable. But for the average person, when they turn on their computer and they're not logged in as a user, it's more likely that what they want to do in the display manager is turn off the computer.

AnErrupTion:
I think a better solution would be to include an option in the configuration to change which function keys to use for shutdown and reboot. What do you think, @lolicon0930?

lolicon0930:
@AnErrupTion Ok, no problem.

AnErrupTion:
@lolicon0930 If you mean if it's the default value, then yes it is.

lolicon0930:
Then why does the explanation say that ly only accepts values ​​from 0 to 8?

AnErrupTion:
That's interesting, I never actually noticed this! It does only accept values from 0 to 8, however I have no idea when this little mistake slipped in 😅

lolicon0930:
Start with an initial commit.

readme.md
```diff
@@ -163,7 +163,7 @@ On Arch Linux, the example .xinitrc (/etc/X11/xinit/xinitrc) starts like this:
 
 ## Tips
 The numlock and capslock state is printed in the top-right corner.
-Use the F1 and F2 keys to respectively shutdown and reboot.
```
AnErrupTion, line 166: I think the default keys should be kept as F1/F2 since we now have the option to configure them.

AnErrupTion (changes requested):
That's the last change I'm requesting, otherwise it's good to go! Thanks for your contribution!

PR #500, opened 2023-04-18
Adding Gentoo-specific intallation step

ToxyFlog1627:
On Gentoo (both my daily driver and VM I reproduced it in) agettys don't get disabled with `rc-update del agetty.ttyN` as they are started by inittab. So to change that, you need to comment out desired line in `/etc/inittab`.
I suggest adding that to documentation for others, like me.

Also, I didn't find it in Gentoo's repos. Is it named not ly? If it isn't in repo, I can try adding it.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -112,6 +112,18 @@ If you choose a tty that already has a login/getty running (has a basic login pr
 # rc-update del agetty.tty2
 ```
 
+On Gentoo if previous command didn't work (there isn't such service), agetty can be disabled in `/etc/inittab` by commenting it out:
+
+```
+...
+# TERMINALS
+#x1:12345:respawn:/sbin/agetty 38400 console linux
+c1:12345:respawn:/sbin/agetty --noclear 38400 tty1 linux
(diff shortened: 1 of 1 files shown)


mazunki:
@ToxyFlog1627 I found some on [Zugaina](https://gpo.zugaina.org/x11-misc/ly), but both of the versioned releases rely on live git repos, which is frowned upon by Gentoo under versioned releases.

I'm currently making an ebuild myself, but I'm lacking versioned releases/tags for `testoaster`, `dragonfail` and `configator`. It'd be very much appreciated if @nullgemm could tag the tip of the current head with some version, so we could anchor to a known version.

Until then, live ebuild should work fine.

ToxyFlog1627:
I saw those, but it would have been better, if it had been in official repository.
Then keep up your work.
Also, did you have issues with making it work as ones I added to readme here? I reproduced it twice in the vm, but want to confirm.

mazunki:
I still haven't compiled nor run it. I am running on s6, though, so my input will not really be useful for you :P

PR #536, opened 2023-08-19
Backup existing config when installing

brunorro:
When installing, checks if `${DESTDIR}/etc/ly/config.ini` exists and if it does, copies it to `${DESTDIR}/etc/ly/config.ini.old`


AnErrupTion (approved):
LGTM

PR #543, opened 2023-09-13
bigclock: attempt to use the same format as the small clock..

ilikenwf:
It will truncate non numeric characters, of course, but this works in my testing...

Let me know if I got the size wrong or anything but I prefer 12 hour formatted time with the leading 0.


SpaghettiBorgar:
Shouldn't this rather be its own option? Might as well make the clock arbitrary width then.

ilikenwf:
Oh perhaps... in my use case I just wanted a quick fix to get the desired look...and out of laziness did it this way, haha.

AnErrupTion:
This PR is superseded by #563 as (hopefully) a less hacky fix.

PR #545, opened 2023-09-23
Add option to change enter key behaviour

goshawk22:
Currently, if you press the enter key after entering your username, it will try to authenticate, even if you haven't entered a password. This adds the option to change the default behaviour so that pressing the enter key while on the username field moves the cursor to the password field. By default the current behaviour is not changed.

I am not used to programming in C so I hope all the changes are acceptable :)

### DIFF
--- a/res/config.ini
+++ b/res/config.ini
@@ -75,6 +75,10 @@
 # File in which to save and load the default desktop and login
 #save_file = /etc/ly/save
 
+# If true, pressing enter will login regardless of current input box. If false, pressing enter will only login if the current
+# input box is for the password. Otherwise it will go to next input box.
+#enter_is_login = true
+
 
 # Remove power management command hints
 #hide_key_hints = false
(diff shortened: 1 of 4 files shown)


AnErrupTion:
@goshawk22 I'm not sure how pertinent this change is. You can already navigate to the password input box using either the arrow keys (and now with the Vim key bindings as well, provided you enable them in the config file), and a user may want to login without a password as well. Is there any justification for this change, or is it just a matter of personal preference? No offense of course, I'm just trying to understand!

goshawk22:
It was mostly a matter of personal preference so I completely understand if you don't want to include it.

AnErrupTion:
Right. I don't think such changes are warranted in the source code directly, but perhaps this behavior could be customizable in the future. No promises though. And sorry again for taking so much time to respond to your PR!

goshawk22:
No worries!

PR #550, opened 2023-10-06
Use `XDG_CONFIG_HOME` for config path instead of default config file path(in `/etc`)

unrealapex:
Prefer using  `XDG_CONFIG_HOME/ly/.config.ini` over `/etc/ly/config.ini` to adhere with XDG specifications and allow config file to be edited without sudo permissions.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -121,7 +121,7 @@ $ make
 # ln -s /etc/sv/ly /var/service/
 ```
 
-By default, ly will run on tty2. To change the tty it must be set in `/etc/ly/config.ini`
+By default, ly will run on tty2. To change the tty it must be set in `~/.config/ly/config.ini`
 
 You should as well disable your existing display manager service if needed, e.g.:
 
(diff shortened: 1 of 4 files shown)


awebeer256:
Isn't the display manager an inherently system-level thing?

unrealapex:
It is, however, I think that most user installed programs should have their configuration files edited in `~/.config/`.

awebeer256:
What does `~` expand to? The display manager runs before any user is logged in.

This is why [system-level configuration is supposed to go in `/etc`](https://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch03s07.html).

unrealapex:
That's a valid point, perhaps it should stay in `/etc`. To answer your question, I created a cronjob:
`crontab -e`
```bash
@reboot ~/homedir
```
`~/homedir`(`/home/aaron/homedir`)
```bash
#!/bin/bash
echo ~ > ~/home
```
```bash
$ systemctl reboot
...
$ # login in as aaron(only other user is root)
$ cat home
/home/aaron
```
I am not 100% sure that the job is running before login though. Perhaps try on your system and see if results are similar.

awebeer256:
> To answer your question, I created a cronjob

Unless I'm very much mistaken, that doesn't prove anything. Cron jobs run as the user whose crontab they're in, so in your example `~` will expand to the home directory of whichever user you ran `crontab -e` as.

My point in asking was to remind you that the display manager is (conceptually at least; I dunno about technically) run as "the system" instead of as any one user *of* the system.

unrealapex:
Thank you for the good analogy. Your argument make me realize how illogical not using a system directory is. Most display managers are configured this way. I should have taken a look into that before submitting this PR. I appreciate the time you took to provide constructive criticism and helping point me in the right direction. Keep up the good work!

Issue #529, opened 2023-07-25
No animation after plymouth-quit-wait.service


Jzavrk:
Using systemd.
Ly works fine but does not animate.

Fixed by disabling splash screen or by adding drop-in file:

```
[Service]
ExecStartPre=-/usr/bin/sleep 1
```

The later solution makes transition not smooth at all.

PR #558, opened 2023-11-11
Fix animation not displaying after window change

### LINKED ISSUE #556: Animation is displayed only after daemon restart
Ly doesn't show animation on startup. 
To show animation I need to login and restart daemon via systemctl
`systemctl restart ly`
Then Ly will start with animation.

I use Arcolinux, Ly version 0.6.0
In /etc/ly/config.ini were changed only two lines:
```ini
animate = true
animation = 1
```

SpaghettiBorgar:
Volatile window changes on startup seem to cause the animation to not show up sometimes.
Simply restarting the animation when the termbox window changed, mitigates it.

Should fix #556 and #529.

### DIFF
--- a/src/draw.c
+++ b/src/draw.c
@@ -776,17 +776,18 @@ static void doom(struct term_buf* term_buf)
 		{0x2588, 8, 4}, // white
 	};
 
+	if ((term_buf->width != term_buf->init_width) || (term_buf->height != term_buf->init_height))
+	{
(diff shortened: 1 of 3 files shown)


AnErrupTion:
Closing since it got fixed in the Zig rewrite. To compensate, I'll tag your account when 1.0.0 is gong to be released for making the original patch.

PR #563, opened 2023-12-18
Added Bigclock configuration

catdotjs:
## Big clock configuration
Just like clock, big clock can be configured. Added new config.ini entry called "bigclock_format".

## Added - as bigclock symbol
We gotta re-write that bigclock.h file. Whatever is being done there right now is downright cursed.

### DIFF
--- a/res/config.ini
+++ b/res/config.ini
@@ -12,6 +12,9 @@
 # enable/disable big clock
 #bigclock = true
 
+# Big clock formatting (see strftime specification) [allowed symbols 1-9,-,:]
+#bigclock_format = %c
+
 # The character used to mask the password
 #asterisk = *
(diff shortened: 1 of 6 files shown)


catdotjs:
I think my implementation can be improved. Specially the reading Unicode characters from config.

AnErrupTion:
@catdotjs Are you willing to update your PR for the Zig rewrite? Since I also have a couple things to say about some of the changes, I think it's best if you close this PR and create a new one. I'll wait for your answer, though (if any).

catdotjs:
I'm not really experienced with zig but the changes I've done at the C are rather trivial so shouldn't be a much of an issue to port if you wish.

PR #564, opened 2023-12-22
Add support for animations after terminal is resized.

wr7:
## Issue
On my laptop, the TTYs will occasionally have an incorrect initial size. This causes animations to only appear  ~50% of the time.
## Recreating the issue
This issue can easily be simulated by running the `make run` target and resizing the terminal window.
## Current behavior
- If the host terminal is resized, `ly` will appear as if no animation is enabled. 
- If the terminal height is `3` and the `matrix` animation is enabled, `ly` will crash.
## New behavior
- The enabled animation will be restarted/reinitialized when the terminal is resized.
- `ly` won't crash when the terminal height is `3` and the `matrix` animation is enabled.

### DIFF
--- a/src/draw.c
+++ b/src/draw.c
@@ -619,11 +619,16 @@ void position_input(
 	password->visible_len = len;
 }
 
-static void doom_init(struct term_buf* buf)
+static void doom_reset(struct term_buf* buf)
 {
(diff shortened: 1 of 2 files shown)


wr7:
Duplicate of #558

PR #569, opened 2024-01-18
Fix warnings and possible overflows in various buffers

kevr:
Hi fairyglade.

This patchset simply allows for the worst-case scenario when snprintfing to buffers, which also removes a warning during compilation.

C strings use a null terminating character, `'\0`, to mark the end of a string for functions that support string inputs without a given size. So, they are always 1 byte bigger than the string that they hold.

### DIFF
--- a/src/login.c
+++ b/src/login.c
@@ -501,8 +501,8 @@ void auth(
 {
 	int ok;
 
-    char tty_id [3];
-    snprintf(tty_id, 3, "%d", config.tty);
+    char tty_id [4];
+    snprintf(tty_id, 4, "%d", config.tty);
 
     // Add XDG environment variables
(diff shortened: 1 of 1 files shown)


AnErrupTion:
Hey there! Would you kindly update ths PR for the Zig rewrite please? Don't worry, it's not hard: in fact, you only have to do that change for the TTY ID now, because it assumes the ID is <=99, even though it's a byte. However, you don't have to worry about the null terminator, because that's handled by the Zig functions:

https://github.com/fairyglade/ly/blob/0586f3424a0e90477fbfa1d0372ded7ee2c0ed0a/src/auth.zig#L24-L25

AnErrupTion:
I've gone ahead and implemented the changes left in master, closing.

Issue #421, opened 2022-08-09
ly crash on artix w/ s6


thinkoid:
I am tinkering with running `ly` with `s6/s6-rc` on artix. When I attempt to run l through the `s6-rc` script it crashes because `tb_init` fails to open `/dev/tty`, leaving the functions tables in `termbox` uninitialized and the program crashes upon touching that data. A simple test around the invocation of `tb_init` in `main` would suffice to avoid a segfault.

The underlying problem remains, i.e., why `tb_init` cannot open `/dev/tty`. I am confused about it and I don't think I understand what ly's expectations are in terms of what tty's run under `getty`, etc. At the moment, my system runs with 7 consoles (7 instances of `getty`) from tty1 to tty7. IIUC, `/dev/tty` is the console of the current process ... and the `s6-supervise` process that spawns `ly` when the service is up-ed has no tty attached, which explains why `/dev/tty` cannot be opened? How does or how should this work? I expected that as soon as the service is launched I would be greeted by the ly interface in ... tty1? Is that correct? Or tty2? Any help is greatly appreciated, thanks.

Cavernosa:
Sorry if i misunderstood, but I think you need to set the tty in the init script. It's how it's done in the [systemd script](https://github.com/fairyglade/ly/blob/master/res/ly.service)

thinkoid:
> Sorry if i misunderstand, but I think you need to set the tty in the init script. It's how it's done in the [systemd script](https://github.com/fairyglade/ly/blob/master/res/ly.service)

From my reading of the source code I think what you are referring to is the TTYPath variable there, set to /dev/tty2? I am not sure about the meaning of those variables as they probably refer to how systemd prepares to run ly.

My reading of the code is that the initialization of termbox library (tb_init) is done by using /dev/tty (which translates to the tty of the current process, i.e., the process that spawns/launches ly). This /dev/tty string is hardcoded in the source. Also, the peculiarity of s6 makes that the launching process, s6-supervise, has no tty. So, this fails to initialize termbox and is the source of the segfault described initially.

Even if I change /dev/tty to an existing tty, like /dev/tty1 or /dev/tty3, it still fails because the code it further assumes that the envar TERM is defined which, by default, is not defined in the environ for the launching s6-supervise process. Now, I could hardcode all these but that is not really how it should run, is it?

[...]

Issue #344, opened 2022-03-24
Make ly compatible with the s6 init system?


qazerito:
Is there a way to enable the ly service on s6, currently it is outputting "failed to access console".

nullgemm:
I never tried s6 but it sure is an interesting project, supporting it as an init system would be amazing

jackghg:
and dinit , openrc, runit

anoraktrend:
Mkinit

thinkoid:
FWIW, I am currently running ly on my Artix with s6+s6-rc, as a service. Not entirely and fully tested but it seems that the basics are there. I am using this repo, [ly-s6](https://github.com/thinkoid/ly-s6). Basically, make sure that getty is not running on tty2 (remove the file named tty2 from getty/contents.d and reload the database). Then, either bring up the service from a terminal, or touch a file named ly in the default/contents.d directory (and, obviously reload) to have it run at the next boot. Again, not fully tested, YMMV. Also, notice that the script is a bash script because I barely started to scratch the surface of execline (the tool of choice in s6 world). Also, it may be the case that making this service a longrun service is pbb not the right choice. It most likely should be a oneshot but I am not sure. HTH

AnErrupTion:
Implemented in master, closing.

PR #580, opened 2024-02-01
Added s6 Support.

TerminalJunki:
- Added service file for starting ly with s6, as well as sufficient instructions to configure it properly. 
- Tested on a fresh Artix Linux vm and my own workstation, so there shouldn't be any issues. 
- Also added option in makefile to install all necessary service files.

#344
#421


AnErrupTion:
Technically "merged" in master. I've added both userbook and you as co-authors, just in case you were two different persons.

TerminalJunki:
Those are both me, userbook is the name of the user on my system. idk how that happened. Anyways thanks!

Issue #278, opened 2021-03-19
locale settings ignored


Phundrak:
When I launch any session from ly (tested in my case with a shell session and Awesome), the locale set in `/etc/locale.conf` is not respected and all locales are set to `en_US.UTF-8`.
```
$ locale
LANG=en_US.UTF-8
LC_CTYPE="en_US.UTF-8"
LC_NUMERIC="en_US.UTF-8"
LC_TIME="en_US.UTF-8"
LC_COLLATE="en_US.UTF-8"
LC_MONETARY="en_US.UTF-8"
LC_MESSAGES="en_US.UTF-8"
LC_PAPER="en_US.UTF-8"
LC_NAME="en_US.UTF-8"
LC_ADDRESS="en_US.UTF-8"
LC_TELEPHONE="en_US.UTF-8"
LC_MEASUREMENT="en_US.UTF-8"
LC_IDENTIFICATION="en_US.UTF-8"
LC_ALL=
```
However, if I manually create a new session from any available TTY (such as TTY1), I get the correct output as set in my locale configuration, be it in the TTY console or in a graphical session started with `startx`.
```
$ locale
LANG=en_DK.UTF-8
LC_CTYPE="en_DK.UTF-8"
LC_NUMERIC=fr_FR.UTF-8
LC_TIME=fr_FR.UTF-8
LC_COLLATE=C
LC_MONETARY=fr_FR.UTF-8
LC_MESSAGES="en_DK.UTF-8"
LC_PAPER=fr_FR.UTF-8
LC_NAME=fr_FR.UTF-8
LC_ADDRESS=fr_FR.UTF-8
LC_TELEPHONE=fr_FR.UTF-8
LC_MEASUREMENT=fr_FR.UTF-8
LC_IDENTIFICATION=fr_FR.UTF-8
LC_ALL=
```
For the record, here is my `/etc/locale.conf` file:
```
LANG=en_DK.UTF-8
LC_COLLATE=C
LC_NAME=fr_FR.UTF-8
LC_NUMERIC=fr_FR.UTF-8
LC_IDENTIFICATION=fr_FR.UTF-8
LC_TELEPHONE=fr_FR.UTF-8
LC_MONETARY=fr_FR.UTF-8
LC_PAPER=fr_FR.UTF-8
LC_ADDRESS=fr_FR.UTF-8
LC_TIME=fr_FR.UTF-8
LC_MEASUREMENT=fr_FR.UTF-8
```
[...]

PR #581, opened 2024-02-12
Do not force LANG env var on login

xelantro:
Regarding issue #278 
Is there any reason we have to set the LANG env var? 
At least on arch systems this causes the custom LC_* configs form /etc/locale.conf to not be applied (/etc/profile.d/locale.sh `if [ -z "$LANG" ]; then`)

### DIFF
--- a/src/login.c
+++ b/src/login.c
@@ -220,11 +220,6 @@ void env_init(struct passwd* pwd)
 	setenv("SHELL", pwd->pw_shell, 1);
 	setenv("USER", pwd->pw_name, 1);
 	setenv("LOGNAME", pwd->pw_name, 1);
-
-	if (lang)
-	{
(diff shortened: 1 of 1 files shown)


AnErrupTion:
Closing since it got fixed in the Zig rewrite. To compensate, I'll tag your account when 1.0.0 is gong to be released for making the original patch.

xelantro:
@AnErrupTion don't worry about it, I just changed a single line :)
I'm just happy when it's finally fixed :heart:

AnErrupTion:
> @AnErrupTion don't worry about it, I just changed a single line :) I'm just happy when it's finally fixed ❤️

I'd feel bad not to credit you even if it's just a single line change! You have no idea how many of those are actually useful 👀

PR #590, opened 2024-03-28
Make Ctrl+Backspace clear input text.

Turu-Tamas:
I wanted to be able to clear the input text with ctrl+backspace instead of ctrl+u.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -280,6 +280,7 @@ int main(int argc, char** argv)
 			case TB_KEY_CTRL_C:
 				run = false;
 				break;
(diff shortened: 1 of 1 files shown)


AnErrupTion:
Ctrl+Backspace, as well as Ctrl+H, aren't standard key bindings for removing a whole line of text in terminals, so I can't accept this PR as it feels (and looks like it is) a personal preference. However, I'm not against the idea of (potentially) having fully configurable key bindings, so if you (@Turu-Tamas) want, create an issue for that!

PR #596, opened 2024-04-15
Zig Rewrite

Kawaii-Ash:
It works on my machine! (ish)


src/animations/Matrix.zig
```diff
@@ -66,102 +66,94 @@ pub fn realloc(self: *Matrix) !void {
...
-                    if (y > self.terminal_buffer.height) break;
 
-                    // Find the head of this column
-                    tail = y;
-                    seg_length = 0;
-                    while (y <= self.terminal_buffer.height and dot.value != ' ' and dot.value != -1) {
+                    // Find the head of this col
+                    tail = i;
+                    seg_len = 0;
+                    while (i <= buf_height and (dot.value != ' ' and dot.value != -1)) {
```
AnErrupTion, line 110: Parentheses are useless here.

src/animations/Matrix.zig
```diff
@@ -66,102 +66,94 @@ pub fn realloc(self: *Matrix) !void {
...
                         dot.is_head = false;
                         if (MID_SCROLL_CHANGE) {
-                            const random = self.terminal_buffer.random.int(i16);
-                            if (@mod(random, 8) == 0) dot.value = @mod(random, MAX_CODEPOINT) + MIN_CODEPOINT;
+                            const randint = self.terminal_buffer.random.int(i16);
+                            if (@mod(randint, 8) == 0)
+                                dot.value = @mod(randint, MAX_CODEPOINT) + MIN_CODEPOINT;
```
AnErrupTion, line 115: For consistency with other code, put the code on the same line as the condition.

src/animations/Matrix.zig
```diff
@@ -66,102 +66,94 @@ pub fn realloc(self: *Matrix) !void {
...
-                    self.dots[y * self.terminal_buffer.width + x].value = @mod(random, MAX_CODEPOINT) + MIN_CODEPOINT;
-                    self.dots[y * self.terminal_buffer.width + x].is_head = true;
+                    const randint = self.terminal_buffer.random.int(i16);
+                    dot.value = @mod(randint, MAX_CODEPOINT) + MIN_CODEPOINT;
+                    dot.is_head = true;
 
-                    if (seg_length > line.length or !first_column) {
-                        self.dots[tail * self.terminal_buffer.width + x].value = ' ';
-                        self.dots[x].value = -1;
+                    if (seg_len > self.lines[j].length or !first_col) {
```
AnErrupTion, line 132: Why don't you use the `line` vairable here?

src/animations/Matrix.zig
```diff
@@ -66,102 +66,94 @@ pub fn realloc(self: *Matrix) !void {
...
+                        self.dots[buf_width * tail + j].value = ' ';
+                        self.dots[j].value = -1;
                     }
-                    first_column = false;
+                    first_col = false;
+                    i += 1;
```
AnErrupTion, line 137: Please use the `while (i <= buf_height) : (i += 1)` form for consistency.

src/auth.zig
```diff
@@ -141,60 +133,80 @@ pub fn authenticate(
...
+    if (shared_err.readError()) |err| return err;
 }
 
-fn initEnv(allocator: Allocator, pwd: *interop.passwd, path: []const u8) !void {
-    const term = interop.getenv("TERM");
-    const lang = interop.getenv("LANG");
+fn initEnv(allocator: Allocator, pwd: *interop.passwd, path: ?[]const u8) !void {
+    const term_env = std.os.getenv("TERM");
+
+    std.c.environ[0] = null;
```
AnErrupTion, line 194: Could you precise why is that needed please?

src/auth.zig
```diff
@@ -226,28 +243,24 @@ fn loginConv(
     const message_count: u32 = @intCast(num_msg);
     const messages = msg.?;
 
-    const response = login_conv_allocator.alloc(interop.pam.pam_response, message_count) catch return interop.pam.PAM_BUF_ERR;
-    defer login_conv_allocator.free(response);
+    var allocator = std.heap.c_allocator;
```
AnErrupTion, line 246: I think this can be made `const`.

src/auth.zig
```diff
@@ -266,28 +286,211 @@ fn resetTerminal(allocator: Allocator, shell: [*:0]const u8, term_reset_cmd: []c
...
+    _ = std.os.waitpid(pid, 0);
+}
+
+fn getFreeDisplay() !u8 {
+    var buf: [15]u8 = undefined;
+    var i: u8 = 0;
+    while (i < 200) : (i += 1) {
+        const xlock = try std.fmt.bufPrint(&buf, "/tmp/.X{d}-lock", .{i});
+        std.os.access(xlock, std.os.F_OK) catch break;
+    }
```
AnErrupTion, line 305: Can't we just iterate over all top-level files in /tmp and check which ones start with `.X` and end with `-lock`? That seems like it'd be much faster than the current method of doing things.

src/auth.zig
```diff
@@ -266,28 +286,211 @@ fn resetTerminal(allocator: Allocator, shell: [*:0]const u8, term_reset_cmd: []c
...
+    return i;
+}
+
+fn getXPid(display_num: u8) !i32 {
+    var buf: [15]u8 = undefined;
+    const file_name = try std.fmt.bufPrint(&buf, "/tmp/.X{d}-lock", .{display_num});
+    const file = try std.fs.openFileAbsolute(file_name, std.fs.File.OpenFlags{});
```
AnErrupTion, line 312: You can use `.{}` instead of `std.fs.File.OpenFlags{}`.

src/auth.zig
```diff
@@ -266,28 +286,211 @@ fn resetTerminal(allocator: Allocator, shell: [*:0]const u8, term_reset_cmd: []c
...
+
+    // Trim trailing slashes
+    var i = xauth_dir.len - 1;
+    while (xauth_dir[i] == '/') i -= 1;
+    const trimmed_xauth_dir = xauth_dir[0 .. i + 1];
+
+    var buf: [256]u8 = undefined;
+    const xauthority: [:0]u8 = try std.fmt.bufPrintZ(&buf, "{s}/{s}", .{ trimmed_xauth_dir, xauth_file });
+    const createFlags = std.fs.File.CreateFlags{};
+    const file = try std.fs.createFileAbsolute(xauthority, createFlags);
```
AnErrupTion, line 367: Same here. (You cna inline the flags.)

src/auth.zig
```diff
@@ -266,28 +286,211 @@ fn resetTerminal(allocator: Allocator, shell: [*:0]const u8, term_reset_cmd: []c
...
+fn xauth(display_name: [:0]u8, shell: [*:0]const u8, pw_dir: [*:0]const u8, xauth_cmd: []const u8, mcookie_cmd: []const u8) !void {
+    var pwd_buf: [100]u8 = undefined;
+    var pwd: [:0]u8 = try std.fmt.bufPrintZ(&pwd_buf, "{s}", .{pw_dir});
+
+    const xauthority = try createXauthFile(pwd);
+    _ = interop.setenv("XAUTHORITY", xauthority, 1);
+    _ = interop.setenv("DISPLAY", display_name, 1);
+
+    const pid = try std.os.fork();
+
```
AnErrupTion, line 382: Remove this empty line for consistency with other code using `fork()`. (Yes, that's an *extremely* minor nitpick :D)

src/auth.zig
```diff
@@ -266,28 +286,211 @@ fn resetTerminal(allocator: Allocator, shell: [*:0]const u8, term_reset_cmd: []c
...
+    var cmd_buffer: [1024]u8 = undefined;
 
     const cmd_str = try std.fmt.bufPrintZ(&cmd_buffer, "{s} {s}", .{ wayland_cmd, desktop_cmd });
     _ = interop.execl(shell, shell, "-c", cmd_str.ptr, @as([*c]const u8, 0));
 }
 
+fn executeX11Cmd(shell: [*:0]const u8, pw_dir: [*:0]const u8, config: Config, desktop_cmd: []const u8, vt: []const u8) !void {
+    const display_num = try getFreeDisplay();
+    var buf: [5]u8 = undefined;
+    var display_name: [:0]u8 = try std.fmt.bufPrintZ(&buf, ":{d}", .{display_num});
```
AnErrupTion, line 403: The explicit type shouldn't be needed here.

src/auth.zig
```diff
@@ -266,28 +286,211 @@ fn resetTerminal(allocator: Allocator, shell: [*:0]const u8, term_reset_cmd: []c
...
+
+    var ok: c_int = undefined;
+    var xcb: ?*interop.xcb.xcb_connection_t = null;
+    while (ok != 0) {
+        xcb = interop.xcb.xcb_connect(null, null);
+        ok = interop.xcb.xcb_connection_has_error(xcb);
+        _ = std.c.kill(pid, 0);
+        if (std.c._errno().* == interop.ESRCH and ok != 0) {
+            return;
+        }
```
AnErrupTion, line 422: For consistency, put the `return;` statement on the same line as the condition.

src/auth.zig
```diff
@@ -266,28 +286,211 @@ fn resetTerminal(allocator: Allocator, shell: [*:0]const u8, term_reset_cmd: []c
...
+    }
+
+    // X Server detaches from the process.
+    // Pid can be fetched from /tmp/X{d}.lock
```
AnErrupTion, line 426: Should be PID, not Pid. :)

src/auth.zig
```diff
@@ -266,28 +286,211 @@ fn resetTerminal(allocator: Allocator, shell: [*:0]const u8, term_reset_cmd: []c
...
 }
 
+fn addUtmpEntry(entry: *Utmp, username: [*:0]const u8, pid: c_int) !void {
+    entry.ut_type = utmp.USER_PROCESS;
+    entry.ut_pid = pid;
+
+    var buf: [4096]u8 = undefined;
+    const ttyname = try std.os.getFdPath(0, &buf);
+
+    var ttyname_buf: [32]u8 = std.mem.zeroes([32]u8);
```
AnErrupTion, line 459: Why are you using `std.mem.zeroes()` here?

src/auth.zig
```diff
@@ -266,28 +286,211 @@ fn resetTerminal(allocator: Allocator, shell: [*:0]const u8, term_reset_cmd: []c
...
+    _ = try std.fmt.bufPrint(&ttyname_buf, "{s}", .{ttyname["/dev/".len..]});
+
+    entry.ut_line = ttyname_buf;
+    entry.ut_id = ttyname_buf["tty".len..7].*;
+
+    var username_buf: [32]u8 = std.mem.zeroes([32]u8);
```
AnErrupTion, line 465: Same here.

src/auth.zig
```diff
@@ -266,28 +286,211 @@ fn resetTerminal(allocator: Allocator, shell: [*:0]const u8, term_reset_cmd: []c
...
+    _ = try std.fmt.bufPrint(&username_buf, "{s}", .{username});
+
+    entry.ut_user = username_buf;
+
+    entry.ut_host = std.mem.zeroes([256]u8);
```
AnErrupTion, line 470: Same here. (Could be split into a variable in order to use `undefined`.)

src/auth.zig
```diff
@@ -266,28 +286,211 @@ fn resetTerminal(allocator: Allocator, shell: [*:0]const u8, term_reset_cmd: []c
...
+    entry.ut_addr_v6[0] = 0;
+
+    utmp.setutent();
+    _ = utmp.pututline(entry);
+}
+
+fn removeUtmpEntry(entry: *Utmp) void {
+    entry.ut_type = utmp.DEAD_PROCESS;
+    entry.ut_line = std.mem.zeroes([32]u8);
+    entry.ut_user = std.mem.zeroes([32]u8);
```
AnErrupTion, line 488: Same here (and same note as above).

src/auth.zig
```diff
@@ -266,28 +286,211 @@ fn resetTerminal(allocator: Allocator, shell: [*:0]const u8, term_reset_cmd: []c
...
+    _ = std.c.gettimeofday(&tv, null);
+
+    entry.ut_tv = .{
+        .tv_sec = @intCast(tv.tv_sec),
+        .tv_usec = @intCast(tv.tv_usec),
+    };
+    entry.ut_addr_v6[0] = 0;
+
+    utmp.setutent();
+    _ = utmp.pututline(entry);
```
AnErrupTion, line 482: Why isn't there `utmp.endutent();`?

AnErrupTion (changes requested):
Thanks for your amazing contribution! There are just.... "a few" things I'll need you to fix or explain.

changelog.md
```diff
@@ -0,0 +1,45 @@
...
+## Save File
+
+The save file is now in .ini format.
+
+Example:
+
+```ini
+user = ash
+session_index = 0
+```
```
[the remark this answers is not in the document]
Kawaii-Ash: It isn't in the C version as far as I'm aware. I thought it just had the user on the first line and session on the 2nd. (no key-value pairs).

src/tui/components/Desktop.zig
```diff
@@ -1,37 +1,35 @@
...
-    Desktop_Entry: struct {
-        Exec: []const u8,
-        Name: []const u8,
-    },
+const DesktopEntry = struct {
+    Exec: []const u8 = "",
+    Name: []const u8 = "",
 };
 
+pub const Entry = struct { Desktop_Entry: DesktopEntry = DesktopEntry{} };
```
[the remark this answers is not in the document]
Kawaii-Ash: Yes. zigini requires each field to either have a default value or be optional.

src/main.zig
```diff
@@ -11,12 +11,20 @@ const TerminalBuffer = @import("tui/TerminalBuffer.zig");
...
+const utils = @import("tui/utils.zig");
 
+const Ini = ini.Ini;
 const termbox = interop.termbox;
 
-const LY_VERSION = "1.0.0";
+pub fn signalHandler(i: c_int) callconv(.C) void {
+    termbox.tb_shutdown();
+    std.c.exit(i);
+}
```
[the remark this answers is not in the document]
Kawaii-Ash: Yes.

src/main.zig
```diff
@@ -98,79 +108,90 @@ pub fn main() !void {
...
+        .mask = std.os.empty_sigset,
+        .flags = 0,
+    };
+    try std.os.sigaction(std.os.SIG.TERM, &act, null);
+
     _ = termbox.tb_select_output_mode(termbox.TB_OUTPUT_NORMAL);
     termbox.tb_clear();
 
+    // we need this to reset it after auth.
+    const orig_tios = try std.os.tcgetattr(std.os.STDIN_FILENO);
```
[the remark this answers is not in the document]
Kawaii-Ash: It's for storing the termios struct after it's set by tb_init. 
Later, it is used after authentication (which now happens in a child process) so the ui doesn't break. 
Calling tb_shutdown and tb_init would result in undesired flashes if authentication failed and no session was started.

Perhaps "tb_termios" would be better.

changelog.md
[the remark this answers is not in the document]
AnErrupTion: Ah yeah, I thought it talked about the config in general, I misread. My bad.

src/main.zig
```diff
@@ -458,63 +512,127 @@ pub fn main() !void {
...
-                ) catch {
-                    has_error = true;
+                _ = std.os.waitpid(session_pid, 0);
+
+                var auth_err = shared_err.readError();
+                if (auth_err) |err| {
                     auth_fails += 1;
                     active_input = .password;
+                    info_line = getAuthErrorMsg(err, lang);
+                    if (config.clear_password or err != error.PamAuthError) password.clear();
```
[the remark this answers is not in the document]
Kawaii-Ash: Normally, the password would be cleared after authentication succeeded.
I believe this has similar behaviour. If the user puts in a wrong password, the input is cleared if the user specified it in the config. 
This avoids the undesirable situation where some error happened after a session ended, then you're in ly with your password not cleared. 
Moving authentication into a child process fixed a few issues, but made things such as this more complicated.

src/interop.zig
```diff
@@ -10,10 +10,14 @@ pub const pam = @cImport({
     @cInclude("security/pam_appl.h");
 });
 
-pub const c_size = u64;
-pub const c_uid = u32;
-pub const c_gid = u32;
-pub const c_time = c_long;
```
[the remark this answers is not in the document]
Kawaii-Ash: Creating aliases for types made the code less clear imo.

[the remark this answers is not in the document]
AnErrupTion: I think it better reflects the C functions being ported over.

src/main.zig
```diff
@@ -98,79 +108,90 @@ pub fn main() !void {
...
 
-    // Load last saved username and desktop selection, if any
-    if (config.ly.load) load_last_saved: {
-        var file = std.fs.openFileAbsolute(config.ly.save_file, .{}) catch break :load_last_saved;
-        defer file.close();
-
-        const reader = file.reader();
-        const username_length = try reader.readIntLittle(u64);
+    var active_input = config.default_input;
+    var vi_mode: ViMode = if (config.vi_mode) .normal else .insert;
```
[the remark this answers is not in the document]
Kawaii-Ash: At the time, i was considering adding a replace mode. But decided against it.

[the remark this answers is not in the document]
AnErrupTion: So, what do we do? Do we keep the enum or do we use a boolean instead?

src/auth.zig
AnErrupTion: Given that the most common scenario is having an `.X0-lock` file, I mark this as resolved since it'll be fast in those scenarios.

src/main.zig
```diff
@@ -98,79 +108,90 @@ pub fn main() !void {
...
+            active_input = .password;
         }
 
-        if (current_desktop < desktop.environments.items.len) desktop.current = current_desktop;
+        if (save.session_index) |session_index| {
+            if (session_index < desktop.environments.items.len) desktop.current = session_index;
+        }
     }
 
-    var active_input = if (config.ly.default_input == .login and login.text.items.len != login.end) .password else config.ly.default_input;
```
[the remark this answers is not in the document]
Kawaii-Ash: I don't think it was working as intended.
Current behaviour is: if there's a saved username, it sets the input to the password field.

src/auth.zig
Kawaii-Ash: It expects a null terminated array. Setting it to undefined and using bufPrintZ works too though.

src/auth.zig
Kawaii-Ash: because it was taken straight from the C version. I don't see a reason why it shouldn't be there though and it has been added.

src/tui/components/Desktop.zig
```diff
@@ -68,33 +66,31 @@ pub fn position(self: *Desktop, x: u64, y: u64, visible_length: u64) void {
...
-        .entry_buffer = undefined,
+        .entry_ini = null,
         .name = name,
         .xdg_name = name, // TODO
         .cmd = cmd,
         .specifier = switch (display_server) {
             .wayland => "wayland",
             .x11 => "x11",
-            else => "other",
+            else => "",
```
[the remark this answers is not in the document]
Kawaii-Ash: was too lazy to pass a translation through. has been added now though.

src/main.zig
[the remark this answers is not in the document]
AnErrupTion: Right, I didn't see the line that set the active input field to the password.

src/auth.zig
Kawaii-Ash: I don't think it is needed anymore.. everything seems to work fine without it now.

src/auth.zig
Kawaii-Ash: It expects a null terminated array so the first index would need to be set. 
Do you think it's worth changing it to this?:

```zig
var host: [256]u8 = undefined;
host[0] = 0;
```

src/main.zig
```diff
@@ -417,13 +458,26 @@ pub fn main() !void {
...
 
         switch (event.key) {
-            termbox.TB_KEY_F1, termbox.TB_KEY_F2, termbox.TB_KEY_F3, termbox.TB_KEY_F4, termbox.TB_KEY_F5, termbox.TB_KEY_F6, termbox.TB_KEY_F7, termbox.TB_KEY_F8, termbox.TB_KEY_F9, termbox.TB_KEY_F10, termbox.TB_KEY_F11, termbox.TB_KEY_F12 => {
-                if (0xFFFF - event.key + 1 == shutdown_key) {
+            termbox.TB_KEY_ESC => {
+                if (config.vi_mode and vi_mode != .normal) {
+                    vi_mode = .normal;
+                    update = true;
+                }
+            },
```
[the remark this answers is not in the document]
Kawaii-Ash: ESC just changes the mode from insert to normal. The mode is already drawn to the screen and I don't think a keyhint is needed for ESC in this scenario. Anyone using vi mode will definitely be aware.

src/interop.zig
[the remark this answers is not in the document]
Kawaii-Ash: In addition to my first point, does that not also break the Zig style guide?

Quoting from the documentation:
"If `x` is a `type` then `x` should be `TitleCase`"

src/main.zig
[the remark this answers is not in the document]
Kawaii-Ash: I've switched it to a boolean. If anyone wants replace mode, then it can be changed back easily enough. But it wasn't mentioned in the issue for this feature.

src/interop.zig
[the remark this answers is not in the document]
AnErrupTion: I think that's only for structs, not primitive types.

src/auth.zig
AnErrupTion: Yeah, I think it is worth it.

src/interop.zig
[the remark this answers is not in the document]
Kawaii-Ash: ah. I read further along and noticed:
`const primitive_type_alias = f32;`

I'm happy to revert this now.

src/main.zig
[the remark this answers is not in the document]
AnErrupTion: Since it's a known and common keybind, I'll mark this as resolved.

res/lang/en.ini
```diff
@@ -34,12 +34,16 @@ err_user_init = failed to initialize user
 err_user_uid = failed to set user UID
 err_xsessions_dir = failed to find sessions folder
 err_xsessions_open = failed to open sessions folder
+insert = Insert
 login = login
 logout = logged out
+normal = Normal
```
AnErrupTion, line 40: `Insert` and `Normal` should be lowercase, just like everything else.

src/auth.zig
```diff
@@ -306,10 +306,10 @@ fn getFreeDisplay() !u8 {
     return i;
 }
 
-fn getXPid(display_num: u8) !i32 {
+fn getXPID(display_num: u8) !i32 {
```
AnErrupTion, line 300: The function name itself should respect camelCase, so it should be `getXPid`. My nitpick was only for comments, not for the actual function name itself 🙃

src/config/Lang.zig
```diff
@@ -41,10 +41,12 @@ login: []const u8 = "login:",
 logout: []const u8 = "logged out",
 normal: []const u8 = "Normal",
```
AnErrupTion, line 42: It should also be lowercase here.

src/main.zig
```diff
@@ -350,11 +353,10 @@ pub fn main() !void {
...
-                if (info_line.len > 0) {
-                    const info_line_width = try utils.strWidth(info_line);
-                    if (buffer.box_width > info_line_width) {
-                        const x = buffer.box_x + ((buffer.box_width - info_line_width) / 2);
-                        buffer.drawLabel(info_line, x, label_y);
+                if (info_line.width > 0) {
+                    if (buffer.box_width > info_line.width) {
+                        const x = buffer.box_x + ((buffer.box_width - info_line.width) / 2);
+                        buffer.drawLabel(info_line.text, x, label_y);
                     }
```
AnErrupTion: You can now combine both conditions in 1 expression.

src/main.zig
```diff
@@ -158,7 +158,7 @@ pub fn main() !void {
     defer password.deinit();
 
     var active_input = config.default_input;
-    var vi_mode: ViMode = if (config.vi_mode) .normal else .insert;
+    var insert_mode: bool = if (config.vi_mode) false else true;
```
AnErrupTion, line 161: You can just set it to `!config.vi_mode`.

src/main.zig
```diff
@@ -387,7 +387,7 @@ pub fn main() !void {
                 }
 
                 if (config.vi_mode) {
-                    const label_txt = if (vi_mode == .normal) lang.normal else lang.insert;
+                    const label_txt = if (!insert_mode) lang.normal else lang.insert;
```
AnErrupTion, line 390: I think it's better to invert that condition (i.e. `if (insert_mode) lang.insert else lang.normal`).

src/auth.zig
```diff
@@ -107,19 +107,12 @@ pub fn authenticate(allocator: Allocator, config: Config, desktop: Desktop, logi
...
             std.os.exit(1);
         };
 
-        // Reset the XDG environment variables from before
-        setXdgSessionEnv(current_environment.display_server);
-        setXdgEnv(allocator, tty_str, current_environment.xdg_name) catch |e| {
-            shared_err.writeError(e);
-            std.os.exit(1);
-        };
-
```
AnErrupTion: Not a request, but we should watch out for those changes in case anything breaks with the environment variables (e.g. they're not set on someone's machine).

AnErrupTion (changes requested):

AnErrupTion (approved):

AnErrupTion:
Amazing work, thank you!

PR #599, opened 2024-04-21
Zig Rewrite - Zig 0.12.0 and more!

### LINKED ISSUE #598: [zig branch] "error: no field named 'DesktopNames'" on compilation attempt
Linux $HOSTNAME 6.8.5-arch1-1 #1 SMP PREEMPT_DYNAMIC Thu, 11 Apr 2024 01:47:33 +0000 x86_64 GNU/Linux
Shell is tcsh 6.24.12, WM is Sway 1.9

Steps to reproduce:
- `git clone git@github.com:fairyglade/ly.git`
- `cd ly`
- `git checkout zig`
- `zig build` (as specified in readme)

Output of `zig build` is here: [error.log](https://github.com/fairyglade/ly/files/15050012/error.log)

Kawaii-Ash:
# Changes

+ updated to use Zig v0.12.0
+ allocator is no longer needed in `authenticate`
+ replaced some functions in interop to use the Zig equivalent
+ mapped `blank_password` to `clear_password` for compatibility with old config
+ cursor is set to the end of the user input when next focused after loading from a save file
+ `save.ini` is expected to be in the same dir as the config
+ changed save migrator to load from old save format correctly. (loads file from `config.save_file`)
+ some general code improvements

Fixes #598


src/main.zig
```diff
@@ -405,12 +395,8 @@ pub fn main() !void {
...
 
                 draw_lock_state: {
-                    const lock_state = interop.getLockState(config.console_dev) catch |err| {
-                        if (err == error.CannotOpenConsoleDev) {
-                            try info_line.setText(lang.err_console_dev);
-                        } else {
-                            try info_line.setText(lang.err_alloc);
-                        }
+                    const lock_state = interop.getLockState(config.console_dev) catch {
+                        try info_line.setText(lang.err_console_dev);
```
AnErrupTion, line 399: Why was this removed?

src/config/migrator.zig
```diff
@@ -2,29 +2,35 @@ const std = @import("std");
...
-    const save = .{
-        .user = user_buffer,
-        .session_index = session_index,
-    };
+    const new_save_file = std.fs.cwd().createFile(new_path, .{}) catch return save;
+    ini.writeFromStruct(save, new_save_file.writer(), null) catch return save;
 
-    ini.writeFromStruct(save, file.writer(), null) catch return save;
+    // Delete old save file
+    std.fs.deleteFileAbsolute(old_path) catch {};
```
AnErrupTion, line 33: Should we do this? This would execute even if `config.save` is false. And, even if it was always true, we'll save it later on anyway.

src/auth.zig
```diff
@@ -71,124 +54,50 @@ pub fn authenticate(allocator: Allocator, config: Config, desktop: Desktop, logi
...
-
-            while (true) : (index += 1) {
-                const char = shell[index];
-                pwd.pw_shell[index] = char;
-
-                if (char == 0) break;
-            }
-        }
+        pwd.pw_shell = interop.getusershell();
+        interop.endusershell();
```
AnErrupTion, line 69: Use `defer` here for consistency with other code.

src/auth.zig
```diff
@@ -230,22 +179,16 @@ fn setXdgSessionEnv(display_server: enums.DisplayServer) void {
     }, 0);
 }
 
-fn setXdgEnv(allocator: Allocator, tty_str: [:0]u8, desktop_name: []const u8) !void {
+fn setXdgEnv(tty_str: [:0]u8, desktop_name: [:0]const u8, desktop_names: [:0]const u8) !void {
```
AnErrupTion, line 182: Name it `xdg_desktop_name` instead of `desktop_names` for better clarity.

src/auth.zig
```diff
@@ -230,22 +179,16 @@ fn setXdgSessionEnv(display_server: enums.DisplayServer) void {
...
 
-    if (desktop_name.len > 0) {
-        const desktop_name_z = try allocator.dupeZ(u8, desktop_name);
-        defer allocator.free(desktop_name_z);
-
-        _ = interop.setenv("XDG_CURRENT_DESKTOP", desktop_name_z.ptr, 0);
-        _ = interop.setenv("XDG_SESSION_DESKTOP", desktop_name_z.ptr, 0);
-    }
-
+    _ = interop.setenv("XDG_CURRENT_DESKTOP", desktop_names.ptr, 0);
```
AnErrupTion, line 187: We should only set those environment variables if we have an XDG name (so if we aren't launching a TTY/shell for example).

src/auth.zig
```diff
@@ -230,22 +179,16 @@ fn setXdgSessionEnv(display_server: enums.DisplayServer) void {
...
     _ = interop.setenv("XDG_RUNTIME_DIR", uid_str.ptr, 0);
     _ = interop.setenv("XDG_SESSION_CLASS", "user", 0);
     _ = interop.setenv("XDG_SESSION_ID", "1", 0);
+    _ = interop.setenv("XDG_SESSION_DESKTOP", desktop_name.ptr, 0);
```
AnErrupTion, line 191: Should be `desktop_names`, not `desktop_name`.

src/auth.zig
```diff
@@ -379,94 +319,99 @@ fn createXauthFile(pwd: [:0]const u8) ![:0]const u8 {
...
+    const pid = try std.posix.fork();
     if (pid == 0) {
         var cmd_buffer: [1024]u8 = undefined;
-        const cmd_str = std.fmt.bufPrintZ(&cmd_buffer, "{s} add {s} . $({s})", .{ xauth_cmd, display_name, mcookie_cmd }) catch std.os.exit(1);
-        _ = interop.execl(shell, shell, "-c", cmd_str.ptr, @as([*c]const u8, 0));
-        std.os.exit(0);
+        const cmd_str = std.fmt.bufPrintZ(&cmd_buffer, "{s} add {s} . $({s})", .{ xauth_cmd, display_name, mcookie_cmd }) catch std.process.exit(1);
+        const args = [_:null]?[*:0]const u8{ shell, "-c", cmd_str };
+        std.posix.execveZ(shell, &args, std.c.environ) catch {};
+        std.process.exit(1);
```
AnErrupTion, line 342: Why is the exit status code `1` here instead of `0` like before?

src/auth.zig
```diff
@@ -379,94 +319,99 @@ fn createXauthFile(pwd: [:0]const u8) ![:0]const u8 {
...
+    const pid = try std.posix.fork();
     if (pid == 0) {
         var cmd_buffer: [1024]u8 = undefined;
-        const cmd_str = std.fmt.bufPrintZ(&cmd_buffer, "{s} {s} {s}", .{ config.x_cmd, display_name, vt }) catch std.os.exit(1);
-        _ = interop.execl(shell, shell, "-c", cmd_str.ptr, @as([*c]const u8, 0));
-        std.os.exit(0);
+        const cmd_str = std.fmt.bufPrintZ(&cmd_buffer, "{s} {s} {s}", .{ config.x_cmd, display_name, vt }) catch std.process.exit(1);
+        const args = [_:null]?[*:0]const u8{ shell, "-c", cmd_str };
+        std.posix.execveZ(shell, &args, std.c.environ) catch {};
+        std.process.exit(1);
```
AnErrupTion, line 372: Same here.

src/auth.zig
```diff
@@ -379,94 +319,99 @@ fn createXauthFile(pwd: [:0]const u8) ![:0]const u8 {
...
+    xorg_pid = try std.posix.fork();
     if (xorg_pid == 0) {
         var cmd_buffer: [1024]u8 = undefined;
-        const cmd_str = std.fmt.bufPrintZ(&cmd_buffer, "{s} {s}", .{ config.x_cmd_setup, desktop_cmd }) catch std.os.exit(1);
-        _ = interop.execl(shell, shell, "-c", cmd_str.ptr, @as([*c]const u8, 0));
-        std.os.exit(0);
+        const cmd_str = std.fmt.bufPrintZ(&cmd_buffer, "{s} {s}", .{ config.x_cmd_setup, desktop_cmd }) catch std.process.exit(1);
+        const args = [_:null]?[*:0]const u8{ shell, "-c", cmd_str };
+        std.posix.execveZ(shell, &args, std.c.environ) catch {};
+        std.process.exit(1);
```
AnErrupTion, line 395: Same here.

src/auth.zig
```diff
@@ -379,94 +319,99 @@ fn createXauthFile(pwd: [:0]const u8) ![:0]const u8 {
...
     interop.xcb.xcb_disconnect(xcb);
 
-    status = std.c.kill(x_pid, 0);
-    if (std.os.errno(status) != .SRCH) {
-        _ = std.c.kill(x_pid, std.os.SIG.TERM);
-        _ = std.c.waitpid(x_pid, &status, 0);
-    }
-}
+    std.posix.kill(x_pid, 0) catch return;
+
```
AnErrupTion, line 410: Remove space here. :)

src/auth.zig
```diff
@@ -379,94 +319,99 @@ fn createXauthFile(pwd: [:0]const u8) ![:0]const u8 {
...
+    std.posix.kill(x_pid, std.posix.SIG.TERM) catch {};
 
-fn executeShellCmd(shell: [*:0]const u8) void {
-    _ = interop.execl(shell, shell, @as([*c]const u8, 0));
+    var status: c_int = 0;
+    _ = std.c.waitpid(x_pid, &status, 0);
```
AnErrupTion, line 414: Any reason `std.c.waitpid()` is used here instead of `std.posix.waitpid()`?

AnErrupTion (changes requested):

src/main.zig
Kawaii-Ash: The function used to use an allocator. The only error it can return now is `error.CannotOpenConsoleDev`.

src/auth.zig
Kawaii-Ash: The process might not exist or it might need some time to be killed. 
If the process doesn't exist, `std.posix.waitpid` will go to `unreachable`.

src/auth.zig
Kawaii-Ash: Since it replaces the process, the only scenario it should reach that point is if an error occurred.

src/auth.zig
Kawaii-Ash: Taking a quote from the arch wiki: "`XDG_SESSION_DESKTOP` is similar to `XDG_CURRENT_DESKTOP,` but only permits a single string."

`desktop_names` could be a colon separated list, meaning it wouldn't be suitable to set `XDG_SESSION_DESKTOP`.

src/auth.zig
Kawaii-Ash: Consistency with what other code?

src/config/migrator.zig
Kawaii-Ash: that's a good point

src/auth.zig
AnErrupTion: Then we should split `xdg_desktop_names` and take like the first item. But we shouldn't set it to `desktop_name`, because that's often separated with spaces (thus not really a single string).

src/auth.zig
AnErrupTion: What I mean is it'd be consistent, as in using `defer` to deinitialize (or in this case, "end") a structure or anything else, just like across the code base.

Kawaii-Ash: it'd be the only usage of defer where a following operation can't fail.

AnErrupTion: That'd be `errdefer` though, not `defer`. The latter can be used in every situation, as in, the statement will always be executed.

Kawaii-Ash:
```
{
    startusershell();
    getusershell();
    endusershell();
}
```
In this situation, `getusershell()` will not result in an error. so `endusershell()` will always execute afterwards. 

It'd only be consistent with the rest of the codebase if the code was as follows:
```
{
    startusershell();
    defer endusershell();
    try getusershell();
}
```
It is otherwise a meaningless change. 
If changes such as this are requested, an explicit style guide should be written down to make it clear.

src/auth.zig
Kawaii-Ash: `DesktopNames` is not required in the session file and could lead to `XDG_SESSION_DESKTOP` not being set in this case.

I don't see any issues regarding `XDG_SESSION_DESKTOP` being set incorrectly in the C version. 
Keeping the current behaviour seems to be more correct.

Kawaii-Ash: The value for `XDG_SESSION_DESKTOP` could possibly be taken from the name of the session file, but there has been no reported issues about changing it thus far.

AnErrupTion: > `DesktopNames` is not required in the session file and could lead to `XDG_SESSION_DESKTOP` not being set in this case.

Then we could fallback to `desktop_name` if that's the case.

> I don't see any issues regarding `XDG_SESSION_DESKTOP` being set incorrectly in the C version. 
> Keeping the current behaviour seems to be more correct.

Most likely because not many applications check for it, but that doesn't mean it's set correctly.

AnErrupTion (approved):
Thank you for your contribution!

PR #604, opened 2024-05-05
[Zig Rewrite] Fix Authentication

### LINKED ISSUE #602: [zig] Unable to sign in and start desktop environment
Hello,
After the changes implemented in 0.12.0, I was finally able to get the zig rewrite of ly to compile.
I set up the systemd service using the recommended method (`zig build installsystemd`) and made some changes to `/etc/ly/config.ini` (enable an animation, set a clock format, and change the masking text from * to -).
However, upon a reboot, a few things happened:
- I can see ly. This is great!
- There is a stray cursor flashing in the top left of the screen maybe 30 columns over and 20 rows down (but I didn't count). This only goes away when I hit an arrow key to move to the password field or change the selected desktop environment.
- No animation is displayed, which is less than ideal - and the reason I moved away from the main branch (see #592). In addition, the other changes I made to the config file are not visible either.
- Finally, when I actually attempt to enter a password, I am greeted with the following...
`exception (no address available)`
`??:?:?: -x73e4ee88709d in ??? (libc.so.6)`
`Unwind information for 'libc.so.0x73e4ee88709d' was not available, trace may be incomplete`
... displayed atop the ly prompt -- so, I am unable to get to sway or i3 via ly, and I have to switch to a different tty and start it manually.

Additional info:
uname: `Linux my_fqdn 6.8.9-arch1-1 #1 SMP PREEMPT_DYNAMIC Thu, 02 May 2024 17:49:46 +0000 x86_64 GNU/Linux`
latest commit hash in my clone of ly: `629749b3832aaec86929794a527c108e5c1c29d1`
init system: systemd
desktop environments tested: i3 and sway

If there are any other questions I can answer or things I can test, let me know. Thanks!

Kawaii-Ash:
This "should" fix #602 

The issue could be reproduced when pam modules such as `pam_motd.so` were used.
Initializing the allocated memory for the response with zeroes fixed it in my tests.

Update: Also updated zigini to fix a mem leak that could occur

### DIFF
--- a/src/auth.zig
+++ b/src/auth.zig
@@ -207,9 +207,7 @@ fn loginConv(
 
     // Initialise allocated memory to 0
     // This ensures memory can be freed by pam on success
-    for (response) |*r| {
-        r.* = std.mem.zeroes(interop.pam.pam_response);
-    }
+    for (response) |*r| r.* = std.mem.zeroes(interop.pam.pam_response);
 
     var username: ?[:0]u8 = null;
     var password: ?[:0]u8 = null;
--- a/src/auth.zig
+++ b/src/auth.zig
@@ -48,7 +48,7 @@ pub fn authenticate(config: Config, desktop: Desktop, login: [:0]const u8, passw
     status = interop.pam.pam_acct_mgmt(handle, 0);
     if (status != interop.pam.PAM_SUCCESS) return pamDiagnose(status);
 
-    status = interop.pam.pam_setcred(handle, 0);
+    status = interop.pam.pam_setcred(handle, interop.pam.PAM_ESTABLISH_CRED);
     if (status != interop.pam.PAM_SUCCESS) return pamDiagnose(status);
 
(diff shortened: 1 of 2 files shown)


src/auth.zig
```diff
@@ -205,6 +205,12 @@ fn loginConv(
     const allocator = std.heap.c_allocator;
     const response = allocator.alloc(interop.pam.pam_response, message_count) catch return interop.pam.PAM_BUF_ERR;
 
+    // Initialise allocated memory to 0
+    // This ensures memory can be freed by pam on success
+    for (response) |*r| {
+        r.* = std.mem.zeroes(interop.pam.pam_response);
+    }
```
AnErrupTion, line 212: You can probably collapse this into a one-liner.

AnErrupTion (approved):
Works on my machine! Just one small request.

AnErrupTion (approved):
Those changes are WAY too BIG. Thanks!

Issue #605, opened 2024-05-12
DESTDIR and NAME reintroduction


s7effeno:
Would it make sense for this project to bring back DESTDIR and NAME to build.zig from the old makefile? Take a look at [this](https://aur.archlinux.org/cgit/aur.git/commit/?h=ly-git&id=36441c49d7719f642e587e75403e20e809927f7c)

AnErrupTion:
Ah yes, that's absolutely worth bringing back. I completely forgot this existed, to be honest 😅. I'll see what I can do probably next weekend.

Oh and by the way, the version for `ly-git` needs to be updated, since Ly is now tracking version 1.0.0 and not 0.6.0.

s7effeno:
Thanks, would you consider tagging the point regarding the updated version?

AnErrupTion:
What do you mean exactly by that? I plan on fixing a few more bugs at least before shipping 1.0.0.

s7effeno:
The version will  automatically be updated after the new release then, you can see how `pkgver()` gets it. It's actually not recommended for me to manually edit for now.

Issue #490, opened 2023-03-15
Shift tab for the previous input.


RannyArcher:
that will be cool, i think by editing only one line you'll make most of us happy.
i tried to edit the src code but i have troubles when running/compiling the code so it's yours masters 🙇🏼‍♂️

Issue #470, opened 2022-12-15
Custom info text


l3td33r:
A way to define a custom text shown on login screen, something like tuigreet has:
![proxy-image](https://user-images.githubusercontent.com/75885440/207918098-bbb2eea1-452f-471d-811f-238ae8a2e3da.png)

PR #606, opened 2024-05-13
Various bug fixes and small features

Kawaii-Ash:
This PR includes some small features and bug fixes.

Features:
+ Custom info text
+ Shift+Tab for navigation
+ Setting numlock at startup
+ Set info line when authenticating
+ Changed build.zig to reintroduce `DESTDIR` and `NAME`

Bug Fixes:
+ Stray Cursor
+ Integer overflows
+ Cascade
+ Fix xauth command in some non-POSIX compliant shells

Related Issues: 
+ #602 (Stray Cursor)
+ #334 
+ #470 
+ #490
+ #605


build.zig
```diff
@@ -42,18 +36,20 @@ pub fn build(b: *std.Build) void {
...
-    exe.addCSourceFile(.{ .file = .{ .path = "dep/termbox_next/src/memstream.c" }, .flags = &c_args });
-    exe.addCSourceFile(.{ .file = .{ .path = "dep/termbox_next/src/ringbuffer.c" }, .flags = &c_args });
-    exe.addCSourceFile(.{ .file = .{ .path = "dep/termbox_next/src/term.c" }, .flags = &c_args });
-    exe.addCSourceFile(.{ .file = .{ .path = "dep/termbox_next/src/termbox.c" }, .flags = &c_args });
-    exe.addCSourceFile(.{ .file = .{ .path = "dep/termbox_next/src/utf8.c" }, .flags = &c_args });
+    // Only fails with ReleaseSafe, so we'll override it.
+    const translate_c = b.addTranslateC(.{
+        .root_source_file = .{ .path = "include/termbox2.h" },
+        .target = target,
+        .optimize = if (optimize == .ReleaseSafe) .ReleaseFast else optimize,
```
AnErrupTion, line 48: We definitely shouldn't do that. Since this is most likely a Zig compiler bug, let it crash with a runtime error for now if ReleaseSafe is used. (Perhaps it has been fixed in Zig 0.13.0?)

build.zig
```diff
@@ -67,120 +63,104 @@ pub fn build(b: *std.Build) void {
...
+    std.fs.cwd().makePath(lang_path) catch {
+        std.debug.print("warn: {s} already exists as a directory.\n", .{data_directory});
     };
 
     var current_dir = std.fs.cwd();
 
     {
-        var executable_dir = std.fs.openDirAbsolute("/usr/bin", .{}) catch unreachable;
+        const exe_path = try std.fs.path.join(allocator, &[_][]const u8{ dest_directory, "/usr/bin" });
+        std.fs.cwd().makePath(exe_path) catch {};
```
AnErrupTion, line 154: Add a warning message for consistency.

build.zig
```diff
@@ -213,9 +193,94 @@ fn install_ly(install_config: bool) !void {
     }
 
     {
-        var pam_dir = std.fs.openDirAbsolute("/etc/pam.d", .{}) catch unreachable;
+        const pam_path = try std.fs.path.join(allocator, &[_][]const u8{ dest_directory, "/etc/pam.d" });
+        std.fs.cwd().makePath(pam_path) catch {};
```
AnErrupTion, line 197: Same here.

build.zig
```diff
@@ -213,9 +193,94 @@ fn install_ly(install_config: bool) !void {
...
+        return version_str;
+    };
+    var git_describe = std.mem.trim(u8, git_describe_raw, " \n\r");
+    git_describe = std.mem.trimLeft(u8, git_describe, "v");
+
+    switch (std.mem.count(u8, git_describe, "-")) {
+        0 => {
+            if (!std.mem.eql(u8, version_str, git_describe)) {
+                std.debug.print("{s} version '{s}' does not match git tag: '{s}'\n", .{ name, version_str, git_describe });
+                std.process.exit(1);
```
AnErrupTion, line 255: Purposefully crashing the build process is a pretty nuclear option in my opinion. We should just let it return `version_str`.

build.zig
```diff
@@ -213,9 +193,94 @@ fn install_ly(install_config: bool) !void {
...
+            // Untagged development build (e.g. 0.10.0-dev.2025+ecf0050a9).
+            var it = std.mem.splitScalar(u8, git_describe, '-');
+            const tagged_ancestor = std.mem.trimLeft(u8, it.first(), "v");
+            const commit_height = it.next().?;
+            const commit_id = it.next().?;
+
+            const ancestor_ver = try std.SemanticVersion.parse(tagged_ancestor);
+            if (version.order(ancestor_ver) != .gt) {
+                std.debug.print("{s} version '{}' must be greater than tagged ancestor '{}'\n", .{ name, version, ancestor_ver });
+                std.process.exit(1);
```
AnErrupTion, line 269: Same here.

src/interop.zig
```diff
@@ -2,9 +2,7 @@ const std = @import("std");
 const builtin = @import("builtin");
 const Allocator = std.mem.Allocator;
 
-pub const termbox = @cImport({
-    @cInclude("termbox.h");
-});
+pub const termbox = @import("termbox2");
```
AnErrupTion, line 5: Because termbox is imported as a Zig module now, it shouldn't be in `interop.zig`.

src/auth.zig
```diff
@@ -335,24 +327,56 @@ fn createXauthFile(pwd: [:0]const u8) ![:0]const u8 {
...
+    const result = std.posix.waitpid(pid, 0);
+
+    if (result.status != 0) return error.CommandFailed;
+
+    var buf: [32]u8 = undefined;
+    const len = try output.read(&buf);
+    if (len != 32) return error.UnexpectedLength;
+    return buf;
+}
+
```
AnErrupTion, line 358: Is all of this really needed? Can't we just use `std.process` to start the process and forward its output to a buffer?

src/main.zig
```diff
@@ -510,9 +516,18 @@ pub fn main() !void {
...
+                    .login => .session,
+                    .password => .login,
+                };
+
+                update = true;
+            },
             termbox.TB_KEY_ENTER => {
                 if (config.save) save_last_settings: {
-                    var file = std.fs.createFileAbsolute(save_path, .{}) catch break :save_last_settings;
+                    var file = std.fs.cwd().createFile(save_path, .{}) catch break :save_last_settings;
```
AnErrupTion, line 530: I feel like allowing a relative path here is trying to shoot ourselves in the foot. Where exactly would the CWD be? In `/usr/bin`? Perhaps `/etc/ly`? This is too ambiguous, in my opinion.

src/bigclock.zig
```diff
@@ -8,89 +8,89 @@ const termbox = interop.termbox;
...
 
-pub const WIDTH: u64 = 5;
-pub const HEIGHT: u64 = 5;
+pub const WIDTH = 5;
+pub const HEIGHT = 5;
 pub const SIZE = WIDTH * HEIGHT;
 
 // zig fmt: off
-const ZERO = [_]u32{
+const ZERO = [_]u21{
```
AnErrupTion, line 16: I'm not convinced by the space savings of this, given the potential runtime cost. Why was this change made?

src/auth.zig
```diff
@@ -129,15 +128,11 @@ fn startSession(
     try initEnv(pwd, config.path);
 
     // Set the PAM variables
-    const pam_env_vars = interop.pam.pam_getenvlist(handle);
+    const pam_env_vars: ?[*:null]?[*:0]u8 = interop.pam.pam_getenvlist(handle);
+    if (pam_env_vars == null) return error.GetEnvListFailed;
```
AnErrupTion, line 132: You've introduced some errors that aren't handled in `main.getAuthErrorMsg()`. This is one of them.

src/auth.zig
```diff
@@ -335,24 +327,56 @@ fn createXauthFile(pwd: [:0]const u8) ![:0]const u8 {
...
+    }
+
+    const result = std.posix.waitpid(pid, 0);
+
+    if (result.status != 0) return error.CommandFailed;
+
+    var buf: [32]u8 = undefined;
+    const len = try output.read(&buf);
+    if (len != 32) return error.UnexpectedLength;
+    return buf;
```
AnErrupTion, line 356: Same here.

src/auth.zig
```diff
@@ -335,24 +327,56 @@ fn createXauthFile(pwd: [:0]const u8) ![:0]const u8 {
...
-        const cmd_str = std.fmt.bufPrintZ(&cmd_buffer, "{s} add {s} . $({s})", .{ xauth_cmd, display_name, mcookie_cmd }) catch std.process.exit(1);
+        const cmd_str = std.fmt.bufPrintZ(&cmd_buffer, "{s} add {s} . {s}", .{ xauth_cmd, display_name, mcookie_output }) catch std.process.exit(1);
         const args = [_:null]?[*:0]const u8{ shell, "-c", cmd_str };
         std.posix.execveZ(shell, &args, std.c.environ) catch {};
         std.process.exit(1);
     }
 
-    _ = std.posix.waitpid(pid, 0);
+    const status = std.posix.waitpid(pid, 0);
+    if (status.status != 0) return error.XauthFailed;
```
AnErrupTion, line 379: Same here.

src/auth.zig
```diff
@@ -388,7 +412,7 @@ fn executeX11Cmd(shell: [*:0]const u8, pw_dir: [*:0]const u8, config: Config, de
         xcb = interop.xcb.xcb_connect(null, null);
         ok = interop.xcb.xcb_connection_has_error(xcb);
         std.posix.kill(pid, 0) catch |e| {
-            if (e == error.ProcessNotFound and ok != 0) return;
+            if (e == error.ProcessNotFound and ok != 0) return error.XcbConnectionFailed;
```
AnErrupTion, line 415: Same here.

AnErrupTion (changes requested):

src/bigclock.zig
Kawaii-Ash: Since a codepoint is 21 bits, it's the correct type to be using. 
The runtime cost is minimal and I don't think it's worth using the wrong type. 

Are we aiming for slightly better performance at the expense of correctness or the opposite?

build.zig
Kawaii-Ash: I'm not sure completely breaking builds in ReleaseSafe is desired when only a single step needs to be overridden, especially when it is considered the "main release mode". 

Editing the comment above it to read: `// HACK: Only fails with ReleaseSafe, so we'll override it for now`, would be a better indicator to change it in the future.

build.zig
Kawaii-Ash: Stopping the build process in such situations is beneficial, since it's an indicator that `ly_version` is set incorrectly.

src/auth.zig
Kawaii-Ash: It's in a fork child, so you want to avoid heap allocation and use functions in `std.posix`. Alternatively, mcookie could be removed as a dependency and it could be implemented directly in Zig in ~5 lines.

src/main.zig
Kawaii-Ash: The cwd is the directory the current user is in or with systemd for example, u set it with `WorkingDirectory=`.
The other paths, for the lang/config can also be a relative path, so it's consistent. 

At times, I'd create a directory with a config just for testing, then I'd provide a relative path to it, which is a lot faster than providing an absolute path. 

I don't see it being used outside of testing things and I'm not sure it should be taken away since a user would hopefully be aware when they're using a relative path.

src/interop.zig
Kawaii-Ash: The only reason it's imported in such a way is due to the workaround to fix ReleaseSafe builds. Probably not worth changing imports across the codebase due to a workaround.

build.zig
AnErrupTion: I'm more concerned about the safety issues it may create. In ReleaseSafe, safety checks (like overflowing, array index out of bounds, etc...) are kept, but not in ReleaseFast. This could potentially dismiss some bugs in Ly by making it automatically build in ReleaseFast even when ReleaseSafe is selected.

src/interop.zig
AnErrupTion: I don't really understand this... how is the current workaround related to the file in which the module is imported?

src/auth.zig
AnErrupTion: > Alternatively, mcookie could be removed as a dependency and it could be implemented directly in Zig in ~5 lines.

Honestly that doesn't sound like a bad idea. Other than that, I understand the reasons why `std.process` isn't used.

src/bigclock.zig
AnErrupTion: > Are we aiming for slightly better performance at the expense of correctness or the opposite?

Oh right, I just didn't understand why the change was made. If it's for correctness, it's fine (Ly doesn't need to be that fast anyway, though of course we shouldn't make it slower for silly reasons).

src/main.zig
AnErrupTion: > The cwd is the directory the current user is in or with systemd for example, u set it with WorkingDirectory=.

Ah right, I suppose that makes sense (and for other init systems using shell scripts, `cd` should work).

build.zig
Kawaii-Ash: As I understand it:
The rest of the project should still have safety checks outside of that module.
Without it, you're forced to compile the whole project with ReleaseFast/ReleaseSmall, so there would be no safety checks at all. 

This temporary workaround is the best case scenario imo.

src/interop.zig
Kawaii-Ash: Normally, it'd be imported like below, but the build mode can't be overridden in this case. 

```
const termbox = @cImport({
    @cDefine("TB_IMPL", {});
    @cInclude("termbox2.h");
});
```

src/auth.zig
Kawaii-Ash: If it's something u want to consider at some point, I threw this together:

```zig
pub fn mcookie() [32]u8 {
    var buf: [4096]u8 = undefined;
    std.crypto.random.bytes(&buf);

    var out: [16]u8 = undefined;
    Md5.hash(&buf, &out, .{});

    return std.fmt.bytesToHex(&out, .lower);
}
```

build.zig
AnErrupTion: Fair enough. However, we should not just have a comment indicating the hack, but also print a message to potentially warn the user (this may not matter for automated buildbots like for distributions, but it's still better than othing).

Kawaii-Ash: I added a warning.

AnErrupTion (approved):

Issue #532, opened 2023-07-28
Start on all ttys


exoess:
I'm just wondering if this is possible, I can't figure it out and I don't wanna switch back to a different display manager

DinoNuggies7:
I second this

exoess:
thank you lol

DinoNuggies7:
I think there might be a way if you setup every tty to auto login and launch ly-dm

mistermuki:
@AnErrupTion do you mind if i work on this?

AnErrupTion:
@mistermuki I don't mind at all, go ahead!

PR #623, opened 2024-06-19
Support for Multiple TTYs

mookums:
This allows for Ly to run on multiple ttys concurrently. It utilizes [systemd templates](https://www.freedesktop.org/software/systemd/man/latest/systemd.unit.html) to support multiple instances. It also will dynamically change the tty, allowing for the desktop environment to start up correctly.

Not sure if this is exactly what we being looked for in #532 but it is a solution that is fairly simple.

### DIFF
--- a/res/ly@.service
+++ b/res/ly@.service
@@ -1,17 +1,18 @@
 [Unit]
-Description=TUI display manager
+Description=TUI display manager (on %I)
 After=systemd-user-sessions.service plymouth-quit-wait.service
-After=getty@tty2.service
-Conflicts=getty@tty2.service
+After=getty@%I.service
+Conflicts=getty@%I.service
 
 [Service]
 Type=idle
 ExecStart=/usr/bin/ly
 StandardError=journal
 StandardInput=tty
-TTYPath=/dev/tty2
+TTYPath=/dev/%I
 TTYReset=yes
 TTYVHangup=yes
 
 [Install]
-Alias=display-manager.service
+WantedBy=multi-user.target
+# Alias=display-manager.service
--- a/src/interop.zig
+++ b/src/interop.zig
@@ -6,6 +6,10 @@ pub const termbox = @cImport({
     @cInclude("termbox.h");
 });
 
+pub const vt = @cImport({
+    @cInclude("linux/vt.h");
+});
+
 pub const pam = @cImport({
     @cInclude("security/pam_appl.h");
 });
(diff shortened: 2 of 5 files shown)


res/ly@.service
```diff
@@ -1,17 +1,18 @@
...
 StandardInput=tty
-TTYPath=/dev/tty2
+TTYPath=/dev/%I
 TTYReset=yes
 TTYVHangup=yes
 
 [Install]
-Alias=display-manager.service
+WantedBy=multi-user.target
+# Alias=display-manager.service
```
AnErrupTion, line 18: Why was the alias to `display-manager.service` removed?

src/interop.zig
```diff
@@ -6,6 +6,10 @@ pub const termbox = @cImport({
     @cInclude("termbox.h");
 });
 
+pub const vt = @cImport({
+    @cInclude("linux/vt.h");
+});
+
```
AnErrupTion, line 12: Is there any other header file that can be included instead of a Linux-specific one?

res/ly@.service
mookums: If you keep it, it doesn't let you enable the service. (Probably because you can enable the template instance multiple times).  You can try it out with `sudo systemctl enable ly@tty2` and it'll throw this error:
`Failed to enable unit: Cannot alias ly@tty2.service as display-manager.service.`

It isn't needed for Ly though, been running it on tty2 and tty9 since I created this pull request. I think sddm and gdm are the only ones that create a display-manager alias.

src/main.zig
```diff
@@ -526,6 +526,11 @@ pub fn main() !void {
                 defer shared_err.deinit();
 
                 {
+                    // This will set the current active VT as our TTY.
+                    var vtstat: interop.vt.vt_stat = undefined;
+                    _ = std.c.ioctl(std.c.STDIN_FILENO, interop.VT_GETSTATE, &vtstat);
+                    config.tty = @intCast(vtstat.v_active);
```
AnErrupTion, line 532: Side note, this would completely override the TTY option and thus make it useless. I think a better idea would be to make the default value `0`, and if it is, get the TTY number from the current active VT.

src/interop.zig
mookums: I'll take a look this week.

AnErrupTion:
Closed on Codeberg.

PR #626, opened 2024-07-01
Added keybinds to control brightness

tubi16:
Added keybinds to control brightness with brightnessctl. 

F5 to decrease brightness.
f6 to increase brightness.


src/main.zig
```diff
@@ -18,6 +18,9 @@ const Save = @import("config/Save.zig");
 const migrator = @import("config/migrator.zig");
 const SharedError = @import("SharedError.zig");
 const utils = @import("tui/utils.zig");
+const c_import = @cImport({
+    @cInclude("unistd.h");
+});
```
AnErrupTion, line 23: Please include this in `interop.zig` instead, and rename it `unistd`, like so:

```suggestion
const unistd = @cImport({
    @cInclude("unistd.h");
});
```

src/main.zig
```diff
@@ -476,6 +479,16 @@ pub fn main() !void {
...
+                } else if (pressed_key == 5) {
+                    if (c_import.access("/usr/bin/brightnessctl", c_import.X_OK) == 0) {
+                        var brightness = std.ChildProcess.init(&[_][]const u8{ "/usr/bin/brightnessctl", "-q", "s", "5%-" }, allocator);
+                        _ = brightness.spawnAndWait()catch.{};
+                    }
+                } else if (pressed_key == 6) {
+                    if (c_import.access("/usr/bin/brightnessctl", c_import.X_OK) == 0) {
+                        var brightness = std.ChildProcess.init(&[_][]const u8{ "/usr/bin/brightnessctl", "-q", "s", "+5%" }, allocator);
+                        _ = brightness.spawnAndWait()catch.{};
+                    }
```
AnErrupTion, line 491: Please implement the following configuration options, along with their respective default values, in `res/config.ini` and `src/config/Config.zig`:

- `brightness_down_key`, whose default should be `f5` (you can parse it the same way as `shutdown_key` and the others)
- `brightness_up_key`, whose default should be `f6` (same as `brightness_down_key`)
- `brightnessctl`, whose default should be `/usr/bin/brightnessctl`

Please also make the brightness key hints show up on the screen, just like the shutdown & reboot keys for example.

Optionally, you can also implement a `brightness_change` option whose default would be `5%` (as I believe `5%` to be a pretty arbitrary choice), but you don't need to.

AnErrupTion (changes requested):

src/main.zig
tubi16: I am gonna do it.  Should I fork latest changes and then make pull request?

AnErrupTion: No need to, you can do it in this PR directly.

tubi16:
@AnErrupTion can you check?  Is ıt as you wanted?

+ brightnessctl package may wanted as dep.

src/main.zig
```diff
@@ -485,15 +501,19 @@ pub fn main() !void {
...
-                        var brightness = std.ChildProcess.init(&[_][]const u8{ "/usr/bin/brightnessctl", "-q", "s", "+5%" }, allocator);
-                        _ = brightness.spawnAndWait()catch.{};
+                } else if (pressed_key == brightness_up_key) {
+                    if (unistd.access(&config.brightnessctl[0], unistd.X_OK) == 0) {
+                        const brightness_str = std.fmt.allocPrint(allocator, "+{s}%", .{config.brightness_change}) catch unreachable;
+                        defer allocator.free(brightness_str);
+                        var brightness = std.ChildProcess.init(&[_][]const u8{ config.brightnessctl, "-q", "s", brightness_str }, allocator);
+                        _ = brightness.spawnAndWait() catch .{};
                     }
                 }
```
AnErrupTion, line 518: Instead of assuming the allocation will always succeed, it may be better instead to set the info line to the `err_alloc` error and break out of the inner condition, like so (same applies for the brightness down key, too):

```suggestion
                } else if (pressed_key == brightness_up_key and unistd.access(&config.brightnessctl[0], unistd.X_OK) == 0) brightness_change: {
                    const brightness_str = std.fmt.allocPrint(allocator, "+{s}%", .{config.brightness_change}) catch {
                        try info_line.setText(lang.err_alloc);
                        break :brightness_change;
                    };
                    defer allocator.free(brightness_str);
                    var brightness = std.ChildProcess.init(&[_][]const u8{ config.brightnessctl, "-q", "s", brightness_str }, allocator);
                    _ = brightness.spawnAndWait() catch .{};
                }
```

AnErrupTion (changes requested):
Just a few changes needed to be done, otherwise it looks good!

AnErrupTion:
> @AnErrupTion can you check? Is ıt as you wanted?
> 
>     * brightnessctl package may wanted as dep.

It can be added as a runtime dependency in the `README.md` file, sure.

src/main.zig
tubi16: Done.

src/config/Lang.zig
```diff
@@ -2,6 +2,7 @@ authenticating: []const u8 = "authenticating...",
 capslock: []const u8 = "capslock",
 err_alloc: []const u8 = "failed memory allocation",
 err_bounds: []const u8 = "out-of-bounds index",
+err_brightness_change: []const u8 = "failed to change brightness",
```
AnErrupTion, line 5: Please add this in the `res/lang/en.ini` file as well.

AnErrupTion (changes requested):

src/config/Lang.zig
tubi16: Added ^^

AnErrupTion (approved):
Thanks for your changes!

PR #630, opened 2024-07-03
Update `minimum_zig_version`

jinzhongjia:
just add `minimum_zig_version` to zig.build.zon,that mean
This allows the compiler to be informed of the required version before compilation


AnErrupTion (approved):
Thanks for your contribution!

Issue #416, opened 2022-07-29
/etc/ly/xsetup.sh is not invoked afeter login (with Qtile using X)


orencio:
Hi, I've installed git version of ly from Arc Linux AUR repository:

```
$ yay -Qs ly | grep "/ly"
local/ly-git 0.5.3.r64.g0a79883-1
```

I've configured to be invoked `/etc/ly/xsetup.sh` when X windows manager is selected for login:

```
$ grep x_cmd_setup /etc/ly/config.ini
x_cmd_setup = /etc/ly/xsetup.sh
```

The script `/etc/ly/xsetup.sh` exists and is executeble. It must load `/etc/profile`:

```
$ ls -lh /etc/ly/xsetup.sh
-rwxr-xr-x 1 root root 3,0K jul 29 10:54 /etc/ly/xsetup.sh

$ grep -A 3 -E '\*\)' /etc/ly/xsetup.sh
  *) # Plain sh, ksh, and anything we do not know.
    [ -f /etc/profile ] && . /etc/profile
    [ -f $HOME/.profile ] && . $HOME/.profile
    ;;
```

Although `/etc/ly/xsetup.sh` is never executed (I select Qtile with X) but when I select wayland windows manager (Qtile with wayland), the script `/etc/ly/wsetup.sh` is correctly executed.

Why `/etc/ly/xsetup.sh` is not invoked?

Regards.

AnErrupTion:
Can you login to Qtile with X? Like, does it all work? (Except of course the xsetup.sh file)

orencio:
Yes. I login without problem, but the script is not executed.

AnErrupTion:
Can you make sure the file has a shebang (``#!/bin/sh``) and that it is executable? (``chmod +x xsetup.sh``)

orencio:
This is the script:

```
$ ls -lh /etc/ly/xsetup.sh
-rwxr-xr-x 1 root root 3,0K jul 29 10:54 /etc/ly/xsetup.sh

$ head -n 1 /etc/ly/xsetup.sh
#! /bin/sh
```

AnErrupTion:
Don't put a space in ``#! /bin/sh``. Just put ``#!/bin/sh``.

orencio:
Ups! What a mistake.

Sorry, I didn't see the space.

PR #640, opened 2024-07-09
Fix incorrect shebang in default xsetup.sh

liesened:
Apparently the incorrect shebang may cause some issues, see #416


AnErrupTion (approved):
This is a very silly mistake. Thanks for catching it!

PR #651, opened 2024-07-14
Fixed 'std' has no member 'ChildProcess' error when building with zig…

0xNiffin:
Hi, just doing a simple change, otherwise it wouldn't build on Zig 0.13.0 :)


AnErrupTion (approved):
Thank you! This went completely unnoticed!

PR #632, opened 2024-07-03
feat: support zig `0.13.0` and setting default TTY at build time

jinzhongjia:
Feature: 
- Add zig `0.13.0` adaptation

It has not yet been tested whether it can be compiled with 12

PR #653, opened 2024-07-17
fix failure to build on zig 0.13

rsekman:
zig 0.13 [renames std.ChildProcess to std.process.Child](https://ziglang.org/download/0.13.0/release-notes.html#Rename-stdChildProcess-to-stdprocessChild) this was reflected in e4abf79 (#632), but was not backported to dc8d143 (#626). consequently ly would FTBFS on zig 0.13. this commit applies the name change in all necessary places.


AnErrupTion:
Fixed already by #651.

PR #654, opened 2024-07-17
Add support for fingerprint login support

### LINKED ISSUE #227: Support for fingerprint login?
I followed [this guide](https://wiki.archlinux.org/index.php/Fprint) to add fingerprint setup to `/etc/pam.d/ly` but unfortunately it didn't work.

It would be nice to add support for fingerprint login and `fprint` has been a great tool to do that.

JAicewizard:
This is a port of https://github.com/kercre123/ly, hence he is added as co-author.
This PR adds support for fingerprint login, and fixes #227.
I tested this in a single-user setup, by patching this into the arch package. logging in as a non-default user was not tested.

### DIFF
--- a/res/pam.d/ly
+++ b/res/pam.d/ly
@@ -1,5 +1,7 @@
 #%PAM-1.0
 
+auth       sufficient   pam_unix-so try_first_pass likeauth nullok # empty-password will not pass this, but will not fail causing the next line to get executed
+-auth      sufficient   pam_fprintd.so # We do not want to get errors when pam_fprintd.so is not present
 auth       include      login
 account    include      login
 password   include      login
--- a/src/auth.zig
+++ b/src/auth.zig
@@ -112,6 +112,102 @@ pub fn authenticate(config: Config, current_environment: Desktop.Environment, lo
     if (shared_err.readError()) |err| return err;
 }
 
+pub fn fingerprintAuth(config: Config, login: [:0]const u8) !*interop.pam.pam_handle {
+    // Open the PAM session
+    var credentials = [_:null]?[*:0]const u8{ login, "" };
+
+    const conv = interop.pam.pam_conv{
+        .conv = loginConv,
+        .appdata_ptr = @ptrCast(&credentials),
+    };
+    var handle: ?*interop.pam.pam_handle = undefined;
+
+    var status = interop.pam.pam_start(config.service_name.ptr, null, &conv, &handle);
+    if (status != interop.pam.PAM_SUCCESS) return pamDiagnose(status);
+
+    // Do the PAM routine
+    status = interop.pam.pam_authenticate(handle, 0);
+    if (status != interop.pam.PAM_SUCCESS) return pamDiagnose(status);
+
+    status = interop.pam.pam_acct_mgmt(handle, 0);
+    if (status != interop.pam.PAM_SUCCESS) return pamDiagnose(status);
+
+    status = interop.pam.pam_setcred(handle, interop.pam.PAM_ESTABLISH_CRED);
+    if (status != interop.pam.PAM_SUCCESS) return pamDiagnose(status);
+
+    status = interop.pam.pam_open_session(handle, 0);
+    if (status != interop.pam.PAM_SUCCESS) return pamDiagnose(status);
+
+    return handle.?;
+}
+
+pub fn postFingerprintAuth(config: Config, desktop: Desktop, handle: *interop.pam.pam_handle, login: [:0]const u8) !void {
+    var tty_buffer: [2]u8 = undefined;
+    const tty_str = try std.fmt.bufPrintZ(&tty_buffer, "{d}", .{config.tty});
+    const current_environment = desktop.environments.items[desktop.current];
+
+    // Set the XDG environment variables
+    setXdgSessionEnv(current_environment.display_server);
+    try setXdgEnv(tty_str, current_environment.xdg_session_desktop, current_environment.xdg_desktop_names orelse "");
+
+    var pwd: *interop.passwd = undefined;
+    {
+        defer interop.endpwent();
+
+        // Get password structure from username
+        pwd = interop.getpwnam(login.ptr) orelse return error.GetPasswordNameFailed;
+    }
+
+    // Set user shell if it hasn't already been set
+    if (pwd.pw_shell[0] == 0) {
+        interop.setusershell();
+        pwd.pw_shell = interop.getusershell();
+        interop.endusershell();
+    }
+
+    var shared_err = try SharedError.init();
+    defer shared_err.deinit();
+
+    child_pid = try std.posix.fork();
+    if (child_pid == 0) {
+        startSession(config, pwd, handle, current_environment) catch |e| {
+            shared_err.writeError(e);
+            std.process.exit(1);
+        };
+        std.process.exit(0);
+    }
+
+    var entry: Utmp = std.mem.zeroes(Utmp);
+    addUtmpEntry(&entry, pwd.pw_name, child_pid) catch {};
+
+    // If we receive SIGTERM, forward it to child_pid
+    const act = std.posix.Sigaction{
+        .handler = .{ .handler = &sessionSignalHandler },
+        .mask = std.posix.empty_sigset,
+        .flags = 0,
+    };
+    try std.posix.sigaction(std.posix.SIG.TERM, &act, null);
+
+    // Wait for the session to stop
+    _ = std.posix.waitpid(child_pid, 0);
+
+    removeUtmpEntry(&entry);
+
+    try resetTerminal(pwd.pw_shell, config.term_reset_cmd);
+
+    // Close the PAM session
+    var status = interop.pam.pam_close_session(handle, 0);
+    if (status != 0) return pamDiagnose(status);
+
+    status = interop.pam.pam_setcred(handle, interop.pam.PAM_DELETE_CRED);
+    if (status != 0) return pamDiagnose(status);
+
+    status = interop.pam.pam_end(handle, status);
+    if (status != 0) return pamDiagnose(status);
+
+    if (shared_err.readError()) |err| return err;
+}
+
 fn startSession(
     config: Config,
     pwd: *interop.passwd,
--- a/src/main.zig
+++ b/src/main.zig
@@ -38,6 +38,33 @@ pub fn signalHandler(i: c_int) callconv(.C) void {
     std.c.exit(i);
 }
 
+// When setting the currentLogin you must deallocate the previous currentLogin first
+var currentLogin: ?[:0]const u8 = null;
+var asyncPamHandle: ?*interop.pam.pam_handle = null;
+
+fn fingerprintLogin(config: Config, login: [:0]const u8) !void {
+    // TODO: loop if there was an error
+    const pamHandle = try auth.fingerprintAuth(config, login);
+    if (currentLogin != null and !std.mem.eql(u8, std.mem.span(@as([*:0]const u8, currentLogin.?)), login)) {
+        return;
+    }
+    asyncPamHandle = pamHandle;
+}
+
+fn startFingerPrintLogin(allocator: std.mem.Allocator, config: Config, login: Text) !void {
+    if (currentLogin) |clogin| {
+        allocator.free(clogin);
+        currentLogin = null;
+    }
+    const login_text = try allocator.dupeZ(u8, login.text.items);
+    currentLogin = login_text;
+    var handle = try std.Thread.spawn(.{}, fingerprintLogin, .{
+        config,
+        login_text,
+    });
+    handle.detach();
+}
+
 pub fn main() !void {
     var gpa = std.heap.GeneralPurposeAllocator(.{}){};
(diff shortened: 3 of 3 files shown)


JAicewizard:
There turned out to be a couple of issues, one was a fast blinking cursor, which has not been fixed, and one issue where after logging in using fingerprint auth, then logging out, logging back in results in an inability to start sway. I tried reproducing the normal auth flow (not pushed) but nothing really fixes. I might have some time to fix this, but some help would be appreciated!

JAicewizard:
I fixed the issue! I also split the normal auth flow in 2, just like the fingerprint auth flow. This removes a lot of duplicate code, and keeps the 2 flows very much similar.

res/pam.d/ly
```diff
@@ -1,5 +1,7 @@
 #%PAM-1.0
 
+auth       sufficient   pam_unix-so try_first_pass likeauth nullok # empty-password will not pass this, but will not fail causing the next line to get executed
```
AnErrupTion, line 3: Is this a typo: `pam_unix-so`? Shouldn't it be `pam_unix.so`?

I'm also unable to find the argument `likeauth` in https://linux.die.net/man/8/pam_unix. What does it do?

src/auth.zig
```diff
@@ -37,24 +30,41 @@ pub fn authenticate(config: Config, current_environment: Desktop.Environment, lo
...
+        if (status != interop.pam.PAM_SUCCESS) _ = interop.pam.pam_end(handle, status);
+    }
     if (status != interop.pam.PAM_SUCCESS) return pamDiagnose(status);
-    defer status = interop.pam.pam_setcred(handle, interop.pam.PAM_DELETE_CRED);
+    return handle.?;
+}
+
+pub fn finaliseAuth(config: Config, current_environment: Desktop.Environment, handle: ?*interop.pam.pam_handle, login: [:0]const u8) !void {
+    var tty_buffer: [2]u8 = undefined;
+    const tty_str = try std.fmt.bufPrintZ(&tty_buffer, "{d}", .{config.tty});
```
AnErrupTion, line 56: Why was this, alongside setting the XDG environment variables process, deferred to later in the code?

src/auth.zig
```diff
@@ -37,24 +30,41 @@ pub fn authenticate(config: Config, current_environment: Desktop.Environment, lo
...
 
-    status = interop.pam.pam_open_session(handle, 0);
+    // Close the PAM session
+    var status = interop.pam.pam_open_session(handle, 0);
+    defer status = interop.pam.pam_close_session(handle, status);
     if (status != interop.pam.PAM_SUCCESS) return pamDiagnose(status);
-    defer status = interop.pam.pam_close_session(handle, 0);
+    defer status = interop.pam.pam_setcred(handle, interop.pam.PAM_DELETE_CRED);
+    defer status = interop.pam.pam_end(handle, status);
```
AnErrupTion, line 63: With the way you've organized the `defer` statements, `pam_end` is executed first, then `pam_setcred`, etc..., which is incorrect. They should be executed in reverse order.

src/main.zig
```diff
@@ -38,6 +38,33 @@ pub fn signalHandler(i: c_int) callconv(.C) void {
...
+    }
+    const login_text = try allocator.dupeZ(u8, login.text.items);
+    currentLogin = login_text;
+    var handle = try std.Thread.spawn(.{}, fingerprintLogin, .{
+        config,
+        login_text,
+    });
+    handle.detach();
+}
+
```
AnErrupTion, line 67: Please put this in the `auth.zig` file. Also, as I understand, this is a form of automatic login, right? Especially given the `try_first_pass` argument passed to the `pam_unix` module beforehand. If so, please rename the functions to `automaticLogin` and `startAutomaticLogin` accordingly.

src/main.zig
```diff
@@ -652,8 +740,20 @@ pub fn main() !void {
...
+                    .login => {
+                        const shouldFingerprint = switch (event.key) {
+                            termbox.TB_KEY_DELETE => true,
+                            termbox.TB_KEY_BACKSPACE, termbox.TB_KEY_BACKSPACE2 => true,
+                            else => event.ch > 31 and event.ch < 127,
+                        };
+
+                        if (shouldFingerprint) {
+                            try startFingerPrintLogin(allocator, config, login);
+                        }
```
AnErrupTion, line 752: This is a bad idea. If anything, it's better to first attempt fingerprint-based login when pressing Enter before attempting to do regular password-based login. But then this wouldn't do much given that the user could simply authenticate normally without a password and it'd do the same thing (at least, as I understand).

src/main.zig
```diff
@@ -477,9 +507,62 @@ pub fn main() !void {
...
+                _ = termbox.tb_clear();
+                _ = termbox.tb_present();
+
+                update = true;
+
+                var restore_cursor = std.process.Child.init(&[_][]const u8{ "/bin/sh", "-c", config.term_restore_cursor_cmd }, allocator);
+                _ = restore_cursor.spawnAndWait() catch .{};
+            }
+            skipEvent = event_error < 0;
+            if ((event_error < 0 and event_error != -6) or event.type == termbox.TB_EVENT_KEY) break;
```
AnErrupTion, line 563: What is this `-6` here? What does it refer to? You can also use `skipEvent` here instead of `event_error < 0` since you defined it right above.

src/main.zig
```diff
@@ -462,7 +492,7 @@ pub fn main() !void {
 
         var timeout: i32 = -1;
 
-        // Calculate the maximum timeout based on current animations, or the (big) clock. If there's none, we wait for the event indefinitely instead
+        // Calculate the maximum timeout based on current animations, or the (big) clock. If there's none, we wait for 100ms instead
```
AnErrupTion, line 495: Is there any reason not to continue waiting for the event indefinitely?

src/main.zig
```diff
@@ -477,9 +507,62 @@ pub fn main() !void {
...
+                }
+
+                const auth_err = shared_err.readError();
+                if (auth_err) |err| {
+                    try info_line.setText(getAuthErrorMsg(err, lang));
+                } else {
+                    try info_line.setText(lang.logout);
+                }
+                asyncPamHandle = null;
+                try startFingerPrintLogin(allocator, config, login);
```
AnErrupTion, line 551: I don't understand. Is it trying to authenticate again after logging out?

AnErrupTion (changes requested):
The more I look at this PR, the more it looks like just an automatic login feature to me. Unless I missed something, nothing seems to be fingerprint-specific and is just trying to first login at startup (which is fine), but also constantly login afterwards (which is less fine). I believe the latter will burn CPU cycles for not much gain, since the user could simply press Enter after touching their fingerprint reader, or so I assume.

To wrap this up, I have a question: if we were to only apply the changes made to the PAM configuration file, would we be able to still login using a fingerprint reader? I'm not trying to downplay your code additions of course, I'm simply trying to understand if the rest you've added could simply be defined as "convenience features", if you see what I mean.
And, if that's the case, then I think this PR should better be rebranded as adding support for automatic login instead of just fingerprint login.

JAicewizard:
I'm sorry, github does not seem to be sending me emails about anything anymore, im subscribed to this thread but I just didn't get any emails!! (I even searched all my emails, nothing).

res/pam.d/ly
JAicewizard: I can indeed also find a post from 2004 (yes, that old) in the redhat bug tracker about it not being documented. It is used in the example configuration for arch (https://wiki.archlinux.org/title/Fprint). I will look into the possibly typo, does indeed look wrong but I don't have time to look into it right now.

src/main.zig
JAicewizard: Yes it is, if someone logged in with their fingerprint, and logged out, they should be able to login with their fingerprint. This might be a problem for something that is permanently plugged into a computer. We could wait for first user-input after logout? This would avoid causing a loop. Optionally we could even make sure the user currently has their cursor on the password field, to allow users to change configuration without being put back into their environment.

src/main.zig
AnErrupTion: Ideally, you should just make it a constant value, so the need for a comment would disappear.

src/main.zig
AnErrupTion: This would probably be solved by making the UI multi-threaded- but we're not there yet, I suppose. Also, you may want to apply this timeout only when fingerprint/automated login is enabled.

src/main.zig
AnErrupTion: But why is it needed in the first place? Can't they just log back in using their fingerprint like they would've done on their first login?

src/auth.zig
AnErrupTion: I think deferring setting the environment variables could cause some issues later on. It's better to leave that where it was before. Also, what duplicate code would this even introduce?

src/main.zig
AnErrupTion: It's a bad idea because the user might not expect Ly to try logging in at each keystroke. Probably here, spawning a single thread that'd try logging in using the fingerprint is a better idea.

AnErrupTion:
@JAicewizard

> This is true, but I think that restarting a PAM login every few minutes is not so bad. We could also increase the timeout to an hour so that there should be basically no wasted cycles (without removing the feature)

I'm not a huge fan of adding such a long artificial timeout, plus, with the link you provided, wouldn't this only apply to the `pam_fprintd.so` module?

> We are also not doing any expensive computations, the most power-hungry part is probably activating the fingerprint reader, but that is simply the cost of the user enabling it, not something we can do anything about

Yes, but think about a laptop for example: even if it's not doing anything expensive, if even just one core is pegged at 100%, the fans may spin really fast, which makes you wonder what Ly is even doing in the background... and it's not great.

> I don't feel like putting in the effort if you do not want the feature, but if you have the intention of merging this if it gets into a better state I will clean it up.

I do, in fact, want to merge this PR! So if you can improve this PR, then do it! :D

JAicewizard:
> Yes, but think about a laptop for example: even if it's not doing anything expensive, if even just one core is pegged at 100%, the fans may spin really fast, which makes you wonder what Ly is even doing in the background... and it's not great.

I understand the concern, but while in PAM land the CPU should be idle. `pam_fprintd.so` offloads to `fprintd` which waits for a finger to be detected and only then starts to actually perform any computations. This waiting is done using some kind of low-level communication with the reader, `fprind` shows 0% usage when waiting for a finger for me, I hope that alleviates these concerns.

> I do, in fact, want to merge this PR! So if you can improve this PR, then do it! :D

Then I will start working on them when I have time :) I will also do a rebase, just so you know

src/main.zig
JAicewizard: Because we need to start the authentication flow again, after a successful authentication the flow stopped and we need to restart it. But I will first rework this code, maybe its good to look at it again after thatn.

src/main.zig
JAicewizard: Yup got it, agree

src/auth.zig
JAicewizard: I fixed this now without duplicating any code, it just adds more code around background login

src/main.zig
JAicewizard: I dont think we can get all the way there. The problem is that when an authentication is already happening, there is no way to cancel it. So if the user switches username for example, the old auth is still running with no way to cancel it, and thus that thread cannot be used to start a new authentication, and we need to start a new thread. However, I did make it so that it only starts a new authentication when you go to the password field, such that very few threads will be spawned, likely just one. I suspect that only a malicious user would create any significant amount of threads.

src/main.zig
JAicewizard: Reworked the code so it looks nicer, and I made is such that on logout it only restarts when entering the password field.

JAicewizard:
I am not sure why github is saying there are merge conflicts, this branch should be fully up-to-date with upstream except one commit which is just about translations, but I think I resolved most issues you pointed out.

hossamdash:
@AnErrupTion , would be great if work on this can continue, I can volunteer as tester, I think this will make Ly the best login manager for most linux and specifically wayland users right now.

inmymum:
Any updates on this?

JAicewizard:
I am still waiting on response, I can rebase this but I would first like some kind of interaction from @AnErrupTion before I put in a lot of effort into it

inmymum:
Lemurs can do this with pam rules, however it hasn't been developed in years

JAicewizard:
Yes I saw the move, since its summer now I should have some more time to rebase and work on it again

AnErrupTion:
@JAicewizard There may be a better way to handle fingerprint login. For example, if you want 2-factor authentication, Ly needs to add support for that explicitly (it's not just as simple as adding a module to the PAM file, see https://man7.org/linux/man-pages/man3/pam_sm_chauthtok.3.html and https://github.com/google/google-authenticator-libpam). I believe there might also be a way to implement some sort of "unattended" login, like for waiting for a fingerprint to be scanned and verified by a certain PAM module.

JAicewizard:
I may be confused, but that is what this PR does, or do you mean implement the waiting inside ly instead of PAM?

PR #658, opened 2024-07-22
Adjust Doom Animation Speed using config.ini

0xNiffin:
Implemented a way to adjust Doom animation speed through the config file and also changed Config.zig so it stays in alphabetical order :) hope that's ok! 

Cheers 🍻


res/config.ini
```diff
@@ -4,6 +4,9 @@
 # matrix -> CMatrix
 animation = none
 
+# Adjust doom animation speed by adding some delay (in milliseconds) when it draws the flames.
+doom_delay = 0
```
AnErrupTion, line 8: This delay should ideally be generic and also be used in the Matrix animation.

AnErrupTion (changes requested):

res/config.ini
0xNiffin: From the testing I've done this only works well with the Doom animation.

I've implemented it in the Matrix animation as well but it ends up just looking very laggy and clunky by just adding 1ms to it.

Cheers

0xNiffin: Ahhh yes, you're right. I didn't notice that option. My bad. 

Cheers

PR #659, opened 2024-07-22
added animation timeout

elloskelling:
The animations have noticeable CPU load.
If the user is working in the tty, or just doesn't log in for a while, there's no need to burn those cycles.
This makes the animation stop after some time, with the default option preserving current behavior (animate forever).
This adds an extra state on top of 'const animate' to minimize code changes.

NB: This also deinits the animations upon timeout, which is currently safe to do, as we also stop calling draw()
nb: The config parameter uses the '_sec' prefix which is generally nice and informative, but none of the other config parameters do that, so it's inconsistent.
n: I chose u8 for the timeout parameter, assuming that if users don't want to animate forever, they probably want to stop within 4 minutes.


src/config/Config.zig
```diff
@@ -56,3 +56,4 @@ brightness_down_key: []const u8 = "F5",
 brightness_up_key: []const u8 = "F6",
 brightnessctl: []const u8 = "/usr/bin/brightnessctl",
 brightness_change: []const u8 = "10",
+animation_timeout_sec: u8 = 0,
```
AnErrupTion, line 59: I think you should bump it up to a `u12`. This way, you can run the animation for up to 68 minutes before it stops!

...Which is obviously overkill, but I think 4 minutes is pretty limiting.

(Also we were so close to hitting 69 minutes with this one... damn it)

AnErrupTion: @elloskelling Perhaps your timeout was a little too high... :D

res/config.ini
```diff
@@ -4,6 +4,11 @@
 # matrix -> CMatrix
 animation = none
 
+# Stop the animation after some time
+# 0 -> Run forever (default)
+# 1..255 -> Stop the animation after this many seconds
+animation_timeout_sec = 0
```
AnErrupTion, line 10: Please also update the comment here to reflect the changes (the maximum value is no longer 2^8 but 2^12).

AnErrupTion (approved):

Issue #646, opened 2024-07-11
Config being ignored


ColetteDiskette:
I just upgraded to 1.0.0-2 (was pre 1.0 prior) before a restart and now my config file seems to be ignored. Notably for me, the CMatrix animation doesn't play, I don't have the big clock, and passwords don't clear on failed attempts. I checked my `/etc/ly/config.ini` and it all still seems to be properly set.

I'm on Arch Linux, and ly is installed via its extra repo.

de-mirage2:
I also have this issue, rolling back to v0.6.0 seems to fix on Arch Linux

Scary-Guy:
I can confirm this as well, though I'm running the ly command directly from /etc/ly restarting may make it work but I feel like it should work this way as well for testing.

I had to completely redo the file with the config from config.ini.old (not replacing, just copying over individual settings I wanted.  Do not replace!)

Checking the file permissions Root should have all the appropriate permissions (because, root) but also it had execute permission off while the backup had it on.  Changing that had no effect.

This is version 1.0, upgraded from 0.6

moabeat-berlin:
I had the exact same issue. Same parameters here:

* Using the ly package from the Arch Linux extra repository.
* Completely recreated the config file from the new example.
* Config was never picked up, neither on reboot nor on restart of the service.

In the end, I found out, it was my 'asterisk' setting in the config.ini. I had it set to '#', because I liked it better. Unfortunately this seems to cause an error in the parsing of the config file, not reporting any error anywhere.

Further testing turned out, the ini parser treats every '#' as an indicator for a comment for the remainder of the line. For other parameters than 'asterisk' this seems to be fine. But if 'asterisk' is empty, the whole config file is ignored.

ColetteDiskette:
After reading that, I uncommented the `asterisk = *` line in my config and fully restarted my computer. My config is still being ignored. Here is my config, for reference:

```
# Animation enabled/disabled
animate = true 

# The active animation
# 0 -> PSX DOOM fire (default)
# 1 -> CMatrix 
animation = 1

# format string for clock in top right corner (see strftime specification)
#clock = %I:%M %p

# enable/disable big clock
bigclock = true

# The character used to mask the password
asterisk = *

# Erase password input on failure
blank_password = true

#The `fg` and `bg` color settings take a digit 0-8 corresponding to:
#define TB_DEFAULT 0x00
#define TB_BLACK   0x01
#define TB_RED     0x02
#define TB_GREEN   0x03
[...]

PR #661, opened 2024-07-27
Add display of error messages

moabeat-berlin:
Addressing issue #646, where the config file got silently ignored.


res/config.ini
```diff
@@ -49,6 +49,12 @@ bg = 0
 # Foreground color id
 fg = 8
 
+# Background color errors
+error_bg = 0
+
+# Foreground color errors
+error_fg = 258
+
```
AnErrupTion, line 57: Is there any "specification" for 16-bit colors like these? It might be interesting to add a comment referring to it, so users can easily change the colors if they wanted to.

AnErrupTion (commented):
While the changes look fine to me otherwise, I'm not a huge fan of introducing another type (`ErrorMessages`) alongside `InfoLine`. I think it's better if an info line takes a parameter to indicate if it's an error, and if so, render it with the error foreground & background colors.

Also, what I planned to do originally was extend the `InfoLine` to add a list of messages instead of just the current one, so you'd be able to cycle through them in the UI at runtime. But of course, you're not in my head so you couldn't have known. :D

res/config.ini
moabeat-berlin: During development, I tried to set the color to `termbox.TB_RED | termbox.TB_BOLD`, like this is similarly done in the matrix animation [here](https://github.com/fairyglade/ly/blob/e775827c8b18d5300c3300486facbfe2e6377436/src/animations/Matrix.zig#L158C44-L158C78). This ended up with the value of `258` not fitting into an `u8`. Also the [colors in termbox2](https://git.sr.ht/~kolunmi/termbox2-zig/tree/master/item/include/termbox2.h#L248) seem to be in `u16`. Therefore I changed it accordingly.

AnErrupTion: No problem then, but then I think it's wise to add a comment explaining what the default value is (i.e. `TB_RED | TB_BOLD`).

moabeat-berlin:
Well, in the best case, no errors are shown at all. :smile_cat: I would think this way it is more obvious, that something is wrong in or with the config. But in the end, it's your design and decision to adopt it like it is now. For me, it would have been a great help and maybe it can be further improved along the way.

AnErrupTion:
> Well, in the best case, no errors are shown at all. 😸 I would think this way it is more obvious, that something is wrong in or with the config. But in the end, it's your design and decision to adopt it like it is now. For me, it would have been a great help and maybe it can be further improved along the way.

Of course! What you can do is extend `InfoLine` to use an array, similar to what you did in `ErrorMessages`, but for now continue to show only the last message in the UI. I can work on adding the rest myself.

moabeat-berlin:
> Of course! What you can do is extend `InfoLine` to use an array, similar to what you did in `ErrorMessages`, but for now continue to show only the last message in the UI. I can work on adding the rest myself.

Alright, sounds good! I will look into this tomorrow.

src/main.zig
```diff
@@ -128,20 +135,23 @@ pub fn main() !void {
         }
     }
 
-    if (!build_options.enable_x11_support) try info_line.setText(lang.no_x11_support);
+    info_line.error_bg = config.error_bg;
+    info_line.error_fg = config.error_fg;
+
+    if (!build_options.enable_x11_support) try info_line.addError(lang.no_x11_support);
```
AnErrupTion, line 141: I wouldn't consider this an error, but just as a simple information to know.

src/main.zig
```diff
@@ -115,7 +119,10 @@ pub fn main() !void {
             save = save_ini.readFileToStruct(save_path) catch migrator.tryMigrateSaveFile(&user_buf, config.save_file);
         }
     } else {
-        config = config_ini.readFileToStructWithMap(build_options.data_directory ++ "/config.ini", mapped_config_fields) catch Config{};
+        config = config_ini.readFileToStructWithMap(build_options.data_directory ++ "/config.ini", mapped_config_fields) catch _config: {
+            try info_line.addError("Unable to parse config file");
```
AnErrupTion, line 123: Please add this in `src/config/Lang.zig` and in `res/lang/en.ini` as something like `err_config`.

src/main.zig
```diff
@@ -616,11 +626,11 @@ pub fn main() !void {
                 if (auth_err) |err| {
                     auth_fails += 1;
                     active_input = .password;
-                    try info_line.setText(getAuthErrorMsg(err, lang));
+                    info_line.setText(getAuthErrorMsg(err, lang));
```
AnErrupTion, line 629: That's considered an error, so please make use of `addError` here.

src/tui/components/InfoLine.zig
```diff
@@ -2,23 +2,54 @@ const std = @import("std");
...
+    var fg: u16 = buffer.fg;
 
-        buffer.drawLabel(self.text, x, label_y);
+    if (self.error_list.items.len > 0) {
+        text = self.error_list.getLast();
+        bg = self.error_bg;
+        fg = self.error_fg;
     }
+
+    const width: u8 = if (text.len > 0) try utils.strWidth(text) else 0;
```
AnErrupTion, line 46: You should avoid recalculating the width every time since it's pointless. What you can do is create an `ErrorMessage` struct and store the text and its width inside it, instead of storing just the text in the array list.

AnErrupTion (changes requested):
Overall the changes look good, just a few minor nitpicks here and there!

src/main.zig
moabeat-berlin: The issue here is, that the language file is read only _after_ the config file and it depends on the config, which language file is chosen.

src/tui/components/InfoLine.zig
moabeat-berlin: When testing my code, I came across a situation when the error text was not centered. This is the case if an error gets added _before_ the info text is set some time later in the code. Since the error has precedence over the info text, it was using the width of the info text while displaying the (longer) error. Didn't look good. To overcome this, put the with calculation in the draw method, so that it is always correct.

AnErrupTion: Yes, so what I meant is that you add an `ErrorMessage` struct that looks like this:

```zig
const ErrorMessage = struct { width: u8, text: []const u8 };
```

So in `addError()`, you would calculate the width of the error message and store it in the struct. Same for `setText()`, but you can just add a `text_width: u8` field in the `InfoLine` struct for that. Then, in `draw()`, you just select the appropriate width given the text you're drawing.

src/main.zig
AnErrupTion: Oh yeah, I forgot about that. In this case that's fine, however please add a comment explaining that (mandatory) choice. Also, a minor nitpick: `Unable` should be `unable` to match the style of other messages :D

src/tui/components/InfoLine.zig
moabeat-berlin: Alright, done.

src/main.zig
moabeat-berlin: Also done as requested.

AnErrupTion (approved):

PR #673, opened 2024-08-02
Adds Farsi Bigclock

KAYT33N:
Hi
I'll open up what each commit does

### 1. Change config type for bigclock
This used to be a "on/off" boolean and I had to either add a new value for language or change this to enum
So I changed it from boolean to enum with following values
"none (default)", "en", "fa"

### 2. Separates big clock's hard codes from its logic
"bigclock.zig" file was already too big to have another language added to it so I separated the clock chars from the functions
This also makes adding new regional or custom numbers sets as easy as filling the "bigclock.Lang.LocaleChars" struct

### 3. Adds Farsi (fa) to big clock langs
This commit creates "fa.zig" file and registers it in codes

### Screenshot
image

Time was : 01:35 -> ۰۱:۳۵

### DIFF
--- a/res/config.ini
+++ b/res/config.ini
@@ -51,8 +51,10 @@ auth_fails = 10
 # Background color id
 bg = 0x0000
 
-# Enable/disable big clock
-bigclock = false
+# Change the state and language of the big clock
+# none -> Disabled (default)
+# en   -> English
+bigclock = none
 
 # Blank main box background
 # Setting to false will make it transparent
--- a/res/config.ini
+++ b/res/config.ini
@@ -54,6 +54,7 @@ bg = 0x0000
 # Change the state and language of the big clock
 # none -> Disabled (default)
 # en   -> English
+# fa   -> Farsi
 bigclock = none
 
 # Blank main box background
--- a/src/bigclock.zig
+++ b/src/bigclock.zig
@@ -37,8 +37,9 @@ pub fn alphaBlit(x: usize, y: usize, tb_width: usize, tb_height: usize, cells: [
 
 fn toBigNumber(char: u8, bigclock: Bigclock) []const u21 {
     const locale_chars = switch (bigclock) {
-        .fa  => fa.locale_chars,
-        else => en.locale_chars,
+        .fa     => fa.locale_chars,
+        .en     => en.locale_chars,
+        .none   => unreachable,
     };
     return switch (char) {
         '0' => &locale_chars.ZERO,
--- a/src/bigclock.zig
+++ b/src/bigclock.zig
@@ -1,111 +1,22 @@
 const std = @import("std");
-const builtin = @import("builtin");
 const interop = @import("interop.zig");
 const utils = @import("tui/utils.zig");
+const enums = @import("enums.zig");
+const Lang  = @import("bigclock/Lang.zig");
+const en    = @import("bigclock/en.zig");
+const termbox    = interop.termbox;
+const Bigclock   = enums.Bigclock;
+pub const WIDTH  = Lang.WIDTH;
+pub const HEIGHT = Lang.HEIGHT;
+pub const SIZE   = Lang.SIZE;
 
-const termbox = interop.termbox;
-
-const X: u32 = if (builtin.os.tag == .linux or builtin.os.tag.isBSD()) 0x2593 else '#';
-const O: u32 = 0;
-
-pub const WIDTH = 5;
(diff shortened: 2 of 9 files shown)


res/config.ini
```diff
@@ -13,8 +13,11 @@ animation_timeout_sec = 0
 # If null, the clock won't be shown
 clock = null
 
-# Enable/disable big clock
-bigclock = false
+# Change the state and language of the big clock
+# none -> Disabled (default)
+# en   -> English
+# fa   -> Farsi
+bigclock = none
```
AnErrupTion, line 20: Please add this change into the config migrator to ensure backwards compatibility. Essentially, all you need to do is make sure the previous boolean value gets converted into the appropriate enum value (so `false` = `none` and `true` = `en`).

src/bigclock/en.zig
```diff
@@ -0,0 +1,92 @@
+const Lang = @import("Lang.zig");
+
+const LocaleChars = Lang.LocaleChars;
+const X = Lang.X;
+const O = Lang.O;
+
+pub const locale_chars = LocaleChars{
```
AnErrupTion, line 7: Please make sure to keep the ``// zig fmt: off`` and ``// zig fmt: on`` comments here.

src/bigclock/fa.zig
```diff
@@ -0,0 +1,92 @@
+const Lang = @import("Lang.zig");
+
+const LocaleChars = Lang.LocaleChars;
+const X = Lang.X;
+const O = Lang.O;
+
+pub const locale_chars = LocaleChars{
```
AnErrupTion, line 7: Same here.

src/bigclock.zig
```diff
@@ -122,19 +35,23 @@ pub fn alphaBlit(buffer: [*]termbox.tb_cell, x: usize, y: usize, tb_width: usize
     }
 }
 
-fn toBigNumber(char: u8) []const u21 {
+fn toBigNumber(char: u8, bigclock: Bigclock) []const u21 {
+    const locale_chars = switch (bigclock) {
+        .fa  => fa.locale_chars,
+        else => en.locale_chars,
+    };
```
AnErrupTion, line 42: In Zig, if you handle all enum possibilities (here `.en` and `.fa`), you don't have to provide an `else` branch. Please do that here.

AnErrupTion (changes requested):
Overall very nice changes, thank you! Just a couple, small more things needed.

src/bigclock.zig
KAYT33N: Actually it gives compile error without the `else` branch because of the `none` variant

AnErrupTion: @KAYT33N You can simply use `unreachable` in the `.none` branch, since we know we can't reach that branch.

src/config/migrator.zig
```diff
@@ -69,6 +69,21 @@ pub fn configFieldHandler(_: std.mem.Allocator, field: ini.IniField) ?ini.IniFie
...
 
+    if (std.mem.eql(u8, field.key, "bigclock")) {
+        // The option now uses a string (which then gets converted into an enum) instead of an boolean
+        // It also includes the ability to change active bigclock's language
+        var mapped_field = field;
+
+        if (std.mem.eql(u8, field.value, "true"))
+            mapped_field.value = "en";
+        if (std.mem.eql(u8, field.value, "false"))
+            mapped_field.value = "none";
```
AnErrupTion, line 80: You can just use an inlined condition here, something like this:

```zig
mapped_field.value = if (std.mem.eql(u8, field.value, "true")) "en" else "none";
```

Checking for `false` isn't worth it.

KAYT33N: Thought of it but this returns `none` on everything right ?
Even `en` will return `none` cause its not equal `true`

AnErrupTion: Ah yes, my bad. In this case, you can just use an `if/else if` statement instead of just two `if`s. Also, `mapped_config_fields` is set incorrectly: it only gets set if at least 1 field needs to be migrated. So, in this case, it needs to be set in either of the two conditions (if `bigclock` is a boolean), else it shouldn't be set.

AnErrupTion (approved):
LGTM, but please update your branch with the latest changes.

KAYT33N:
@AnErrupTion 
Rebased with master
Thanks for the help on doing this

AnErrupTion:
Thanks for your contribution!

PR #681, opened 2024-08-07
Fix EXE_NAME variable in ly-openrc

DoctorKnowsBetter:
Without the EXE_NAME variable defined, the service cannot be started.


AnErrupTion:
@DoctorKnowsBetter You can simply rename `$EXE_NAME` into `$EXECUTABLE_NAME`. It'll get patched in automatically by the build system, but for some reason I completely forgot to change that one.

PR #683, opened 2024-08-08
Center the environment name.

llc0930:


AnErrupTion:
@llc0930 It certainly makes it look a bit odd since the username & password fields aren't centered either, which is why I do think this is best made a config option.

llc0930:
Modify it after some time.

llc0930:
zig doesn't seem to support optional arguments? So the way I wrote it might be a little ugly.

AnErrupTion (approved):

PR #698, opened 2024-09-30
Update readme.md

widkit:
added zig to dnf install command.


AnErrupTion (approved):
Thanks for the small fix!

Issue #695, opened 2024-09-19
[Bug] Error message of brightness commands displaying at current cursor position


moritz-reinel:
### Pre-requisites

- [X] I have looked for any other duplicate issues

### Ly version

1.1.0-dev.98+022d146

### Observed behavior

When executable for the `brightness_down_cmd` or `brightness_up_cmd` doesn't exist, the stderr output of the subprocess is displayed at the current cursor position

### Expected behavior

Either fail "silently" or add the error to the status bar.

I already noticed the (currently unused) lang item `err_brightness_change` which could be used in the status bar for implementation.

### Steps to reproduce

Set the config option `brightness_down_cmd` to any non-existent executable

### Relevant logs

```shell
                   ┌───────────────────────────────────────────────┐
                   │                                               │
                   │  <        failed to access console         >  │
                   │                                               │
                   │  x11      < Plasma (X11)                   >  │
                   │                                               │
                   │  login    /bin/sh: 1: /usr/bin/brightnessctl: not found
                   │                                               │
                   │  password                                     │
                   │                                               │
                   └───────────────────────────────────────────────┘
```

moritz-reinel:
I have already looked into solving the problem and could open a PR, if appreciated :)

AnErrupTion:
> I have already looked into solving the problem and could open a PR, if appreciated :)

That'd be greatly appreciated!

AnErrupTion:
> I have a similar issue. In my case its an error message about a bluetooth device

That's a different issue. That's due to the TTY logs flowing into Ly's input fields.

PR #699, opened 2024-09-30
Prevent output of brightness/sleep command showing on screen (fixing #695)

moritz-reinel:
This fixes issue #695.

For now, I only implemented the output suppression of the sleep command. Imo, it would make sense to handle the sleep command the same way as the brightness commands (aka. also showing an error message indicating the command failed). I did not implement that yet because it would require adding a lang attribute like `err_sleep` and touching every lang file (to keep consistency in the code).


AnErrupTion (approved):
LGTM. Thanks for your contribution! And indeed, it'd be nice to do this for the sleep command as well.

PR #701, opened 2024-10-08
Add Hyprland to tested DEs

Ch1llyB1lly:
Adds hyprland to the list of tested desktop environments

I have tested ly with different configs with hyprland and has produced no problems


AnErrupTion (approved):
Thanks for testing!

PR #703, opened 2024-10-12
Implement error reporting for sleep command

moritz-reinel:
Added error reporting in case of failure of the sleep command, analogue to the brightness commands


AnErrupTion (approved):

PR #704, opened 2024-10-12
Make language files uniform

moritz-reinel:
# Motivation

During implementation of #703, I noticed that the language files were a bit inconsistent. Some had empty lines for missing translations, others didn't do that.

To make this a bit more uniform, I wrote a script (and applied it) to the current state of lang files.

# Implementation

At first it looks for all the possible "language items" in `Lang.zig`. Then it goes through each <lang>.ini file and saves all the key-value pairs into an array. Finally, the script loops through the items found in Lang.zig again and puts it as `lang-item = <value-found-in-current-file>` or `<empty line>`, overriding the current lang file.

Also I sorted the language items in ASCII order, as it was a bit inconsistent between Lang.zig and the lang files.

PS: At first I made the script in Python. Afterwards, I decided to opt for bash instead, as it should be more available to people using/contributing to ly. (Korn shell wasn't an option unfortunately, as it doesn't support associative arrays)

# Result

With that change, a translator can open Lang.zig and his desired language.ini side by side, then look at empty lines and by matching the line number get the key from Lang.zig and add the translation.

Also after adding a new "language item" (like err_sleep), the script should be run.


AnErrupTion (commented):
@moritz-reinel I like the idea, but I feel like a Python script would be much, much more readable while still being accessible to all contributors. Python is quite universal these days, so I wouldn't see this as a problem.

moritz-reinel:
@AnErrupTion Re-added the python version and did some refactoring. PR would be ready to merge from my side.

Although, thinking more about it, would it be possible/make sense to do that as some "build step" that can be triggered manually?
Would keep the tools needed to zig and not introduce anything.

Issue #708, opened 2024-10-21
[Feature] Delta time for animations


peterc-s:
### Pre-requisites

- [X] I have looked for any other duplicate issues
- [X] I have confirmed the requested feature doesn't exist in the latest version in development

### Wanted behavior

Animations currently speed up and slow down, the fire animation - for example - goes way too fast and occasionally slows down. The username/password cursor blinks so incredibly fast, however the big clock's second counter remains perfectly fine. Would it be possible to incorporate some form of delta time into the animations to make sure they run smoothly?

Lxtharia:
Noticed that too. Would be nice if you could adjust the speed of the animation via the config as well!

peterc-s:
> Noticed that too. Would be nice if you could adjust the speed of the animation via the config as well!

I opened a PR where I added that feature, allows you to set the minimum number of milliseconds between frames in the config file.

PR #709, opened 2024-10-22
added refresh interval in ms for animations

peterc-s:
Added a configuration option that allows the user to set the time between animation frames in milliseconds called `animation_refresh_ms`.

Had to change around the animation code to separate updating the animation buffer and drawing to the terminal buffer a bit, and if there is an animation a new `std.time.Timer` is used to check if enough time has passed to draw with an update.

Sort of what I wanted out of #708

### DIFF
--- a/res/config.ini
+++ b/res/config.ini
@@ -35,6 +35,14 @@
 # matrix -> CMatrix
 animation = none
 
+# The minimum time between animation frames
+# 32 -> ~30fps
+# 16 -> ~60fps
+# 8  -> ~120fps
+# 6  -> ~144fps
+# 4  -> ~240fps
+animation_refresh_ms = 16
+
 # Stop the animation after some time
 # 0 -> Run forever
 # 1..2e12 -> Stop the animation after this many seconds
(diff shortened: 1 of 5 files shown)


peterc-s:
This seems to cause a segfault when actually logging in for the doom animation, and when on the matrix animation, it causes the display to go blank whilst authenticating. May need some help here.

peterc-s:
It seems like some code added to give up the TTY on authentication (commit c7f70ac78f84cdf566d30726a849ee92b95c3c22) and then take it back afterwards caused a segmentation fault with the doom animation and potentially a double free. Temporarily removed that code until I figure out how to fix it.

AClon314:
I've build and this is ok. I set interval to `128` to slow down animation, for keeping error log printed on `ly` at the first time, then let `cmatrix` flush buffer to clean up these error log.

AnErrupTion:
Closed on Codeberg already, plus GitHub PRs won't be merged anymore (see Codeberg migration notice), therefore closing. Sorry for that!

PR #743, opened 2025-04-10
Added image rendering using nearest color matching

Augusto-p:
I added support for displaying images in Ly using its existing color palette.
Each pixel is mapped to the closest available color using Euclidean distance in RGB space.
The image is rendered using Ly’s existing animation system, so no additional complexity is introduced to the rendering process.

This allows background or decorative images to be shown while maintaining the minimalist and efficient nature of the project.

👉 [Create a custom image](https://augusto-p.github.io/ly-Image-maker/)


AnErrupTion:
@Augusto-p Hey, sorry for the late response! Pull requests on GitHub are no longer accepted as the development is now continuing on Codeberg. Sorry!

PR #763, opened 2025-10-05
change(config): add -n flag to prevent accidental backlight disabling

mctaylors:


mctaylors:
didn't saw the migration notice

PR #769, opened 2026-03-29
config: allow waylandsessions and xsessions to be set to null

sabrehagen:
## Description

`waylandsessions` and `xsessions` are currently non-optional string fields, so
there is no clean way to disable session type discovery for users who do not use
Wayland or X11. Setting them to a nonexistent path works but produces log errors
on every startup.

This change makes both fields optional (`?[]const u8`), consistent with other
nullable config fields such as `xinitrc`. Setting either to `null` in `config.ini`
cleanly skips crawling for that session type with no side effects.


AnErrupTion:

PR #770, opened 2026-03-29
config: add shell option to hide the shell session

sabrehagen:
## Description

The shell session is unconditionally added to the session list with no way to
hide it. This is inconsistent with `xinitrc`, which is omitted from the list
when set to `null`.

This change adds a `shell` boolean config option (default `true`). Setting it
to `false` hides the shell session from the list, following the same pattern
as `xinitrc`.


AnErrupTion:

PR #771, opened 2026-03-29
config: add show_tty option to display active TTY in top right corner

sabrehagen:
## Description

When running multiple ly instances across different TTYs there is no way to tell
which TTY a given login screen belongs to at a glance.

This change adds a `show_tty` boolean config option (default `false`) that
displays the active TTY number (e.g. `tty3`) in the top right corner. When the
clock is also enabled the TTY label sits immediately to its right on the same
row. When the clock is disabled it occupies the top right corner on its own.

I'm open to advice from the maintainers on the placement of the TTY label —
positioning it next to the clock is simply my personal preference and it doesn't
need to stay there if a different position is more appropriate.


AnErrupTion:

PR #772, opened 2026-03-29
config: add vt_default_sessions option for per-VT default session

sabrehagen:
## Description

When ly runs on multiple TTYs the saved session is per-user and global — there
is no way to configure a different default session for each TTY. For example, a
user may want TTY1 to default to an X11 session and TTY3 to a custom session.

This change adds a `vt_default_sessions` config option that maps TTY numbers to
session names. The format is a comma-separated list of `vt:name` or
`vt:type:name` entries, where `type` is one of `wayland`, `x11`, `custom`,
`shell`, or `xinitrc`. Specifying a type disambiguates sessions that share the
same name across types (e.g. Ubuntu exists as both a Wayland and X11 session).
Without a type, the first matching session in list order (wayland, then x11,
then custom) is used. When a VT default is configured, it overrides the session
that ly would otherwise restore from the user's last-login save state, ensuring
the intended session is always pre-selected on that TTY regardless of what the
user previously chose. Autologin is unaffected.

Example: `vt_default_sessions = 1:x11:ubuntu,3:custom:my-session`

`findSessionByName` is refactored to delegate to a new `findSessionByNameAndType`
helper that accepts an optional `DisplayServer` filter, keeping the existing
call sites unchanged.


AnErrupTion:

PR #773, opened 2026-04-16
Added option to move the box relative to the screen size

MartorSkull:
## What are the changes about?

Added a new option in the configuration file for moving the box relative to the screen size.

```
box_h_position = 0.5
box_v_position = 0.5
```

The big clock is centered relative to the box. In the cases where it would be outside of the screen, it moves the box to fit in the screen.

## What existing issue does this resolve?

Add more options for personalization

## Examples

Normal usage:
```
box_h_position = 0.15
box_v_position = 0.35
```
image

Clock would be outside of the screen vertically:
```
box_h_position = 0.15
box_v_position = 0.0
```
image

Clock would be outside of the screen horizontally and vertically:
```
box_h_position = 0.0
box_v_position = 0.0
input_len = 3
```
image

## Pre-requisites

- [x] I have tested & confirmed the changes work locally
- [x] I have run `zig fmt` throughout my changes


MartorSkull:
Will move it to codeberg

